PR #98, opened 2015-04-18
Fix typo in handling of no_rewrite_prefixes

tilgovi:
I think this must be a typo.


tilgovi:
If so, a patch release would be greatly appreciated.

ikreymer:
ugh, yeah, it is.. i really need to add a client-side test suite for wombat.js. I'll do another release soon.
In the meantime, can adjust it in the config.yaml for via, the names are arbitrary.. just need to be consistent.

PR #102, opened 2015-05-14
Added change directory command to samples and tests

machawk1:
Without this additional command, the subsequent commands won't run for the naive user.


ikreymer:
ah, cool, thanks for the fix

PR #127, opened 2015-09-10
Correction of "index" plurality

machawk1:


ikreymer:
oops, that was not into master.. i'll just make the change

PR #133, opened 2015-10-16
Handle unicode in linked url.

jcushman:
Hi, Ilya!

We're getting a funky playback for this URL: http://www.ustavni.sud.rs/page/view/139-100028/ustav-republike-srbije

Playback is here: https://perma.cc/E34Q-R44Q?type=source

Seems to be caused by unicode in a link within the page somewhere (`/Storage/Global/Documents/Misc/Информатор_2015.pdf`) which causes an exception in urljoin:

```
  File "/home/vagrant/.virtualenvs/perma/lib/python2.7/site-packages/pywb/rewrite/url_rewriter.py", line 123, in urljoin
    new_url = urlparse.urljoin(orig_url, url)
  File "/usr/lib/python2.7/urlparse.py", line 300, in urljoin
    return urlunparse((scheme, netloc, '/'.join(segments),
UnicodeDecodeError: 'ascii' codec can't decode byte 0xd0 in position 0: ordinal not in range(128)
```

I added something here to catch the UnicodeDecodeError and do something that's apparently reasonable with it (based on how Django handles unicode in URLs). (I'm not 100% sure which python versions `urllib.quote` works for, but it seems to work for 2.7 and 3?)

This seems to allow it to play back without exceptions (although there's still some funky stuff going on with the playback of the left menu).

Thanks!
Jack

### DIFF
--- a/pywb/rewrite/url_rewriter.py
+++ b/pywb/rewrite/url_rewriter.py
@@ -1,4 +1,4 @@
-import urlparse
+import urlparse, urllib
 
 from wburl import WbUrl
 from cookie_rewriter import get_cookie_rewriter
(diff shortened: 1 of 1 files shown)


ikreymer:
Hi Jack,

Hmm, for some reason this does not repro in the latest version as per:

https://webrecorder.io/live/http://www.ustavni.sud.rs/page/view/139-100028/ustav-republike-srbije

But looks like perma is not too far behind, but just in case, can you send me the WARC so I can also test with latest?

jcushman:
Ah ha! Your comparison tipped me off -- we were returning unicode() instead of str() from our cdx server. If we return strings the playback is perfect. Closing this.

Found an issue with handling spaces around href contents while investing this, but I'll file a separate bug. :)

Thanks!

PR #139, opened 2015-10-22
Use webencodings to encode head_insert_str.

jcushman:
Here's a basic implementation of using python-webencodings to handle character encoding.

This correctly causes the head insert to be added to http://www.geocities.jp/koreanlaws/yousiengumi.html using shift_jis encoding instead of falling back on utf-8. (I confirmed this via print statements.)

I've thought of two cases where this might actually matter:

- If the collection name or some other part of wbinfo has unicode characters. (Probably a good idea in general to support foreign-language collection names, right?)
- If the page is in some non-utf8-compatible encoding (utf16? others?) set via response headers.

... but I confess I haven't actually tested either of those.

### DIFF
--- a/pywb/rewrite/rewrite_content.py
+++ b/pywb/rewrite/rewrite_content.py
@@ -1,5 +1,6 @@
 #import chardet
 import pkgutil
+import webencodings
 import yaml
 import re
 
@@ -164,7 +165,7 @@ class RewriteContent:
 
                 if charset:
                     try:
-                        head_insert_str = head_insert_orig.encode(charset)
+                        head_insert_str = webencodings.encode(head_insert_orig, charset)
                     except:
                         pass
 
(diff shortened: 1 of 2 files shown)


jcushman:
(And I guess a third case where it would matter is if you used a custom head_insert with unicode characters.)

ikreymer:
Awesome, thanks! I will see if I can test this out some more..

As a side note, for utf-16, would still need to decode the whole stream. Currently, the rewriting actually treats the output as bytes, which has overall worked surprisingly well for most pages which are ascii compatible -- the rewriter only needs to check for html tags, and urls are rewritten url-encoded. The reason for this is that in the beginning when trying to decode using the html specified encoding, it would often not work (although webencodings may solve this!). Probably ideal would be to decode the stream only if not ascii-compatible, otherwise leave as is

Issue #148, opened 2015-11-23
baseURI override fails in Edge


robertknight:
The `HTMLDocument.baseURI` override fails in Edge. The document.cookie getter in Edge can be retrieved via the standardized method: `Object.getOwnPropertyDescriptor(Document.prototype, 'cookie')` but not via the currently used method of `document.__lookupGetter__('cookie')`

robertknight:
Related - https://developers.google.com/web/updates/2015/04/DOM-attributes-now-on-the-prototype-chain?hl=en

ikreymer:
Relevant changelists: 1484b06da61a886a6517c34f0b0b2d610cb557c1, 977647cea9a9888fd161ec11e7ffe8b64350634e, 520811729924b719746613f3545e3d0fc15c29d1

Issue #147, opened 2015-11-13
Many monkey-patched DOM accessors broken in Safari when live rewriting is enabled


robertknight:
Many DOM accessors that are monkey-patched by `wombat.js` are broken in Safari where fetching the original getter/setter implementations using either `__lookupGetter__` or `Object.getOwnPropertyDescriptor()` does not work. See https://github.com/hypothesis/via/issues/59 and https://bugs.webkit.org/show_bug.cgi?id=140575

---

**Update**: In Safari 9, Object.getOwnPropertyDescriptor() for native DOM properties returns a descriptor object but the getter, setter and configurable bits are incorrect. In particular, the getter and setter are undefined. See the upstream bug at https://bugs.webkit.org/show_bug.cgi?id=49739#c19 for discussion.

robertknight:
See also - http://stackoverflow.com/questions/992461/is-it-possible-to-override-document-cookie-in-webkit

ikreymer:
Thanks, yeah, I think we need to update wombat.js to gracefully degrade for browsers that do not support all the overrides...

robertknight:
This error also appears to be affecting Edge as well: https://github.com/hypothesis/via/issues/55

ikreymer:
Ok, well that's good to know that this is all related.. I was focusing on Chrome and FF, but clearly need to test other browsers, and even if rewriting won't work, it should avoid these errors..

robertknight:
I'm looking into the possibilities for detecting the browser limitations that prevent the current approach to JS rewriting. There don't appear to be any tests that exercise wombat.js' rewriting functionality at present. Is this correct or did I miss something?

ikreymer:
Yes, you are correct, unfortunately, there is not a client-side test suite yet.. it is high on my list of priorities, I would be happy to collaborate on building one.. Do you have any suggestions? I was thinking of using: https://splinter.readthedocs.org/en/latest/ with Selenium..

robertknight:
My main priority at the moment is end-to-end testing of Via, for which I was planning to start with either Selenium or possibly the WebDriver clients (eg. chromedriver) directly. We're considering using Sauce Labs or BrowserStack for cross-browser testing.

In the process of setting that up I'll no doubt learn a lot that might come in useful for testing pywb. I can't promise how much time I'd have to help with that directly.

What is the best documentation available about what `wombat.js` does, how it works etc. besides the code itself?

ikreymer:
[...]

PR #149, opened 2015-11-25
Karma tests for Wombat JS

robertknight:
This adds a set of Karma tests for the client-side JS in pywb and configuration to run them during Travis CI builds.

*In order for the tests to run as part of a Travis CI build, Sauce Labs credentials need to be set via the SAUCE_USERNAME and SAUCE_ACCESS_KEY env vars. These are available free for OSS projects.*

The initial test is a complete integration test for `wombat.js` which works by creating a fresh Window and Document in an iframe for each test, loading wombat.js into it and then performing some tests on content in that iframe.

In future I would strongly suggest that wombat.js be split up into smaller pieces which can be more easily tested on their own.

The tests use Sauce Labs to test against a suite of browsers, they are initially configured
to run against Chrome and Firefox. Safari and Edge are currently failing for the reasons noted in #147 and #148

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,10 @@ python:
 os:
   - linux
 
+cache:
+  directories:
+    - $HOME/.cache/pip
+    - node_modules
 
 sudo: false
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,9 +15,11 @@ install:
   - pip install boto certauth
   - python setup.py -q install
   - pip install coverage pytest-cov coveralls --use-mirrors
+  - npm install
 
 script:
-    python setup.py test
+  - python setup.py test
+  - cd karma-tests && make test
 
 after_success:
     coveralls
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -8,6 +8,16 @@ os:
   - linux
 #  - osx
 
+addons:
+  sauce_connect:
+    username: "robertknight"
+    access_key: "6775e8c2-68fb-4f12-9c12-2bf326d2eba3"
```
robertknight: I have included this key in plain text for now, when this lands on master it should have its own set of credentials added owned by the repository admin and added to .travis.yml using `travis encrypt`

Sauce Labs accounts are free for OSS projects.

ikreymer:
Same here.. best to put into develop (or separate) branch as master is for the 0.10.10 release currently.. thanks for all this!

ikreymer:
Created this: https://github.com/ikreymer/pywb/tree/client-side-tests

ikreymer:
Replaced by #152

PR #150, opened 2015-11-25
Fix Wombat in Safari 9

robertknight:
In Safari 9, Object.getOwnPropertyDescriptor(domObject.prototype, prop)
returns descriptors which have the correct structure but undefined
getters and are marked as unconfigurable.

See https://bugs.webkit.org/show_bug.cgi?id=49739#c19
for details. The getters cannot be retrieved via obj.__lookupGetter__()
either.

 * Resolve the issue by skipping overrides for DOM properties
   where the property is not configurable, or the original getter for
   a property could not be retrieved.

 * Lookup the 'baseURI' property on the correct prototype (Node,
   not document)

This fix also resolves the problem where accesses to document.baseURI
on Edge would fail.

Tests for this require the infrastructure added in https://github.com/ikreymer/pywb/pull/149 . Once the test infrastructure is available I can post a PR that adds Safari to the test suite.

### DIFF
--- a/pywb/static/wombat.js
+++ b/pywb/static/wombat.js
@@ -363,26 +363,31 @@ var wombat_internal = function($wbwindow) {
     }
 
     //============================================
-    // Define custom property
+    // Override a DOM property
     function def_prop(obj, prop, set_func, get_func) {
+        // if the property is marked as non-configurable in the current
+        // browser, skip the override
+        var existingDescriptor = Object.getOwnPropertyDescriptor(obj, prop);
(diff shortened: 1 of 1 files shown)


ikreymer:
Awesome.. can you change this to merge into the develop branch also? (or could have a separate branch for client side tests, just so i can keep track)

ikreymer:
Created this: https://github.com/ikreymer/pywb/tree/client-side-tests

ikreymer:
Superceded by #153

PR #151, opened 2015-11-25
Fix test_rewrite_live.test_example_1

robertknight:
One of the live-rewrite tests fetches 'http://example.com' and had hardcoded assumptions
about the value of the 'Content-Length' header in the response and
the casing of the header.

The expected content length is for the uncompressed content size.
However, the current version of the requests lib sets the
'Accept-Encoding: gzip' header by default and also lowercases
the names of headers in the response.

The proper approach here would be to mock out the HTTP request.

This takes a slightly uglier initial step of performing a direct
request and comparing the X-Archive-Content-Length header
in the proxy's response to that of the original URL.

A related issue is that a `test_loaders` request which also queries example.com
is failing and it looks like it may be for the same reason (compressed vs. non-compressed response)

### DIFF
--- a/pywb/rewrite/test/test_rewrite_live.py
+++ b/pywb/rewrite/test/test_rewrite_live.py
@@ -1,3 +1,5 @@
+import requests
+
 from pywb.rewrite.rewrite_live import LiveRewriter
 from pywb.rewrite.url_rewriter import UrlRewriter
 from pywb.rewrite.wburl import WbUrl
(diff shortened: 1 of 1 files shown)


ikreymer:
Hey, just catching up on all these.. i thought i had fixed this in develop but maybe not..
can you make make all the pull requests against the develop branch, as that is the latest for the 0.11 release?

robertknight:
Thanks @ikreymer - It isn't obvious from the homepage that 'master' is not the current version and that development happens on a separate branch. What is the reason for that, as opposed to doing development directly on master?

ikreymer:
I was trying to follow this idea: http://nvie.com/posts/a-successful-git-branching-model/

ikreymer:
It kind of makes sense.. the latest stable version is on master, less stable on 'develop'.. though I could see other arguments also.., there are tagged versions for releases as well.

ikreymer:
See 2922801b7ca930ed237ed2081b2d84a68b586b10, though agree mock response is a better solution for all the live tests

PR #153, opened 2015-11-26
Fix Wombat in Safari 9

ikreymer:
In Safari 9, Object.getOwnPropertyDescriptor(domObject.prototype, prop)
returns descriptors which have the correct structure but undefined
getters and are marked as unconfigurable.

See https://bugs.webkit.org/show_bug.cgi?id=49739#c19
for details. The getters cannot be retrieved via obj.__lookupGetter__()
either.

 * Resolve the issue by skipping overrides for DOM properties
   where the property is not configurable, or the original getter for
   a property could not be retrieved.

 * Lookup the 'baseURI' property on the correct prototype (Node,
   not document)

This fix also resolves the problem where accesses to document.baseURI
on Edge would fail.

### DIFF
--- a/pywb/static/wombat.js
+++ b/pywb/static/wombat.js
@@ -363,26 +363,31 @@ var wombat_internal = function($wbwindow) {
     }
 
     //============================================
-    // Define custom property
+    // Override a DOM property
     function def_prop(obj, prop, set_func, get_func) {
+        // if the property is marked as non-configurable in the current
+        // browser, skip the override
+        var existingDescriptor = Object.getOwnPropertyDescriptor(obj, prop);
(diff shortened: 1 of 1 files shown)


ikreymer:
Supercedes #150

robertknight:
Thanks for the speedy merge @ikreymer ! I'll check out the client-side-tests branch locally and check the current status under Safari and Edge. Couple of questions:

1) What are the plans for getting a new release of pywb out?

2) Would you have some time to catch up (maybe via IRC or Hangouts) to talk about pywb and Via generally and perhaps also your email to the H dev mailing list a while back about your experiments with netcapsule as an alternative direction to Via.

PR #155, opened 2015-11-26
Enable testing against Safari on Sauce Labs and local browsers when SL credentials are not available

robertknight:
Enable testing against Safari on Sauce Labs

 * Provide a fallback mode in the Karma tests which tests
   against a local browser (defaults to Firefox) if Sauce Labs
   credentials are not set.

   This is useful for local testing for contributors who
   might not have a Sauce Labs account.

 * Add Safari under OS X to the set of Sauce Labs browsers
   that the Karma tests are run against, following the merge
   of the WombatJS fixes for Safari and Edge.

   Although Edge now works under manual testing, automated
   testing against Sauce Labs is not yet working for reasons
   yet to be determined.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,3 @@
+[*.js]
+indent_style=space
+indent_size=4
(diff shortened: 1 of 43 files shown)


robertknight:
I opened this against the wrong branch. Will close and re-create.

PR #156, opened 2015-11-26
Enable testing against Safari on Sauce Labs.

robertknight:
(This PR also includes https://github.com/ikreymer/pywb/pull/154, which is necessary for the tests to pass)

* Provide a fallback mode in the Karma tests which tests
  against a local browser (defaults to Firefox) if Sauce Labs
  credentials are not set.
  This is useful for local testing for contributors who
  might not have a Sauce Labs account.
* Add Safari under OS X to the set of Sauce Labs browsers
  that the Karma tests are run against, following the merge
  of the WombatJS fixes for Safari and Edge.

Although Edge now works under manual testing, automated testing against Edge via Sauce Labs is not yet working for reasons yet to be determined.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,9 @@ python:
 os:
   - linux
 
+addons:
+  sauce_connect: true
+
 cache:
   directories:
     - $HOME/.cache/pip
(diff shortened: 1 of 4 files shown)


robertknight:
^ The checks are failing because the PR is being run against Travis using a local Firefox instance, but Travis is not configured to run Firefox headlessly. I'm currently working on that.

PR #157, opened 2015-11-26
Microsoft Edge compatibility fixes

### LINKED ISSUE #148: baseURI override fails in Edge

robertknight:
(Note: This PR also includes the changes from #156 and #154 , I can either rebase once they are merged or we can just merge this one).

This adds Microsoft Edge to the Sauce Labs test configuration in Karma, fixes a couple of quirks with how the tests run (eg. Edge did not load JS code in an <iframe> if the iframe's main URL failed to load, whereas the other browsers did) and also corrects the object that the baseURI override is installed onto.

When using a version of pywb with these combined fixes, Via works against Edge and Safari.

Fixes #148

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,9 @@ python:
 os:
   - linux
 
+addons:
+  sauce_connect: true
+
 cache:
   directories:
     - $HOME/.cache/pip
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,10 @@ install:
   - pip install coverage pytest-cov coveralls --use-mirrors
   - npm install
 
+before_script:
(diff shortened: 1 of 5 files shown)


ikreymer:
Wow, that's awesome! I'll merge this one instead of the others then.. I've just registered for Sauce labs and will add keys as per: https://docs.travis-ci.com/user/sauce-connect/

PR #215, opened 2017-04-28
Wombat proxies discussion

N0taN3rd:
This PR is to discuss the utilization of the JS Proxy built-in with wombat.js and how it ties in with pywb's html/content rewriting.


ikreymer:
Thanks!
Why is the long list of 'no modify' objects in wombat.js needed? There's no way around it?

N0taN3rd:
@ikreymer I believe it is not be needed as the 'own function' object would suffice to differentiate between the functions needing to bind _$wbwindow_ to the function.
I initially only had the 'no modify' object before adding the 'own function' object which accomplishes the same thing so it is an unneeded artifact of the original exploratory implementation. 
 
The document proxy would need a similar 'own function' object as well see the replay of the angular documentation [page](http://wsdl-docker.cs.odu.edu:8080/heavyJs/20170425050333/https://angular.io/docs/ts/latest/).
The current implementation minus commit 0f76b46 is [deployed](http://wsdl-docker.cs.odu.edu:8080/) and contains two collections [heavyJs](http://wsdl-docker.cs.odu.edu:8080/heavyJs) and [notSoHeavyJs](http://wsdl-docker.cs.odu.edu:8080/notSoHeavyJs). All URI-Ms are listed on the collections page.

ikreymer:
Sorry for the delay, I've started cleanup of the rewriting system, which will need to be done to integrate this better. I think it'd be helpful to separet out the wombat.js changes from the python changes necessary to support this. It seems like the key changes on the python side are:
- Customizing regex rewriting.
- Ability to wrap the text, so custom first buffer + last buffer for the JS rewriter.

Ideally, this would just involve one custom JS regex rewriter which also supports adding first + last buffers. I'm not sure I understand the difference between `firstbuf_proxy2` and `firstbuf_proxy` one exactly...

The new rewriting system is on this branch, and at the moment here is the default configuration:
https://github.com/ikreymer/pywb/blob/ea5c6305e7a6518a59a4256741406a9a1717dbb7/pywb/urlrewrite/rewriter.py

I can imagine there be an alternative rewriter that uses a different JS rewriter to support this, as well as an alternate wombat.js -- I think at least initially would like to support both the old system and the new with some kind of config setting. The new branch is still in flux but will let you know once things are ready.

I can add a WrappingJSRewriter for example that could provide the wrapping support needed here in one class.

N0taN3rd:
@ikreymer 
The key difference between ``firstbuf_proxy2`` and ``firstbuf_proxy`` is that 
``firstbuf_proxy2`` handles the service worker case by accounting for the fact that the globals _WB_wombat_window_proxy, _WB_wombat_document_proxy and WB_wombat_location are not availble in the service worker context and 
an error is thrown when attempting to access them. So if the window globals are there accessing them does not through an error otherwise we are in a worker context. The function definition is seen below whereas the minified version of below is what is used in the python code.
```js
var _____assignFunction_____ = function _____assignFunction_____ (which) {
  let returnValue;
  switch (which) {
    case 'window':
    case 'top':
      try {
        returnValue = _WB_wombat_window_proxy; // if service worker this throughs an error
      } catch(error) {
        returnValue = {};
      }
      break;
    case 'self':
      try {
        returnValue = _WB_wombat_window_proxy;  // if service worker this throughs an error
      } catch(error) {
        returnValue = self;
      }
      break;
    case 'location':
      try {
        returnValue = WB_wombat_location;  // if service worker this throughs an error
      } catch(error) {
        returnValue = {};
      }
      break;
    case 'document':
      try {
        returnValue = _WB_wombat_document_proxy;  // if service worker this throughs an error
      } catch(error) {
        returnValue = {};
      }
      break;
  }
  return returnValue;
}
```
The python rewritting would also need to handle:
Webgl code defined in  script tags
```html
<script id="vs" type="x-shader/x-vertex">
{
let window = _WB_wombat_window_proxy;
let self = _WB_wombat_window_proxy;
let document = _WB_wombat_document_proxy;
let location = WB_wombat_location;
let top = WB_wombat_top;

	uniform sampler2D map;
	uniform float width;
	uniform float height;
	uniform float nearClipping, farClipping;
	uniform float pointSize;
	uniform float zOffset;
	varying vec2 vUv;

	const float XtoZ = 1.11146; // tan( 1.0144686 / 2.0 ) * 2.0;
	const float YtoZ = 0.83359; // tan( 0.7898090 / 2.0 ) * 2.0;

	void main() {
		vUv = vec2( position.x / width, position.y / height );
		vec4 color = texture2D( map, vUv );
		float depth = ( color.r + color.g + color.b ) / 3.0;
		// Projection code by @kcmic
		float z = ( 1.0 - depth ) * (farClipping - nearClipping) + nearClipping;
		vec4 pos = vec4(
			( position.x / width - 0.5 ) * z * XtoZ,
			( position.y / height - 0.5 ) * z * YtoZ,
			- z + zOffset,
			1.0);
		gl_PointSize = pointSize;
		gl_Position = projectionMatrix * modelViewMatrix * pos;
	}


}</script>
```
- http://wsdl-docker.cs.odu.edu:8080/heavyJs/20170426050138/http://void-ii.hi-res.net
- http://wsdl-docker.cs.odu.edu:8080/heavyJs/20170426043243/http://www.moodsmusic.it/
- http://wsdl-docker.cs.odu.edu:8080/heavyJs/20170426005945/http://gemstones.elespacio.net/

As well as script tag used house only JSON
```html
<script id="initial-props" type="application/json" data-reactid=".1nbjvcu9hq8.1.2">
{
let window = _WB_wombat_window_proxy;
let self = _WB_wombat_window_proxy;
let document = _WB_wombat_document_proxy;
let location = WB_wombat_location;
let top = WB_wombat_top;
{"path":"/","assets":{"main":"bundle.js"},"name":"shade","title":"shade","href":"/shade","version":"1.1.0","description":"Mathematically derived gradient explorer","keywords":["React","Design Tool","Gradients","Color"],"links":[{"href":"//github.com/jxnblk/shade","text":"GitHub"}]}

}</script>
```
- http://wsdl-docker.cs.odu.edu:8080/heavyJs/20170425040714/http://jxnblk.com/shade/
- http://wsdl-docker.cs.odu.edu:8080/heavyJs/http://vision-explorer.reactive.ai/#/galaxy?_k=mg4zzz
- http://wsdl-docker.cs.odu.edu:8080/heavyJs/20170426010940/http://radio.garden/live/

N0taN3rd:
@ikreyme are you interested in receiving help in fixing issues, refactoring process, or documentation beyond our continuing discussion?

> I think it'd be helpful to separate out the wombat.js changes from the python changes necessary to support this

I have [documented](https://n0tan3rd.github.io/wombatDocumentationES6) Wombat which is available for [viewing](https://n0tan3rd.github.io/wombatDocumentationES6/identifiers.html) along with the [viewable source code](https://n0tan3rd.github.io/wombatDocumentationES6/source.html) from within the documentation. 

If you are interested in accepting another contribution to pywb/wombat like this let me know otherwise I will take it down.

ikreymer:
@N0taN3rd Sure, that would be great! I wasn't sure about your availability, but if you do have time to contribute to refactoring to integrate into a future release, that would definitely help! And of course would credit you for your contribution. The documentation setup is great!

Right now I am in the process of refactoring the rewriting module to make it more flexible in general. The idea is to able to support custom rewriting options more easily for different scenarios (rewriting, banner-only proxy, rewriting proxy, etc...) as well as different JS rewriting options to support different versions of wombat. I can provide more information about the plan and maybe would make sense to chat next week about a general plan?

N0taN3rd:
@ikreymer  Awesome!!! I do have more time now that the spring session is over and more than happy to help out in any way I can. 

I agree it would make sense to chat about the plan and the only day of the week I know I will not be available for sure is Friday. 
Timewise, I am flexible and have become a night owl as of late (my time EST) so finding a time that works for both of us should not be too hard.

Is there anything specific you wish to be done in the interim? 
One of the things I was thinking about was creating a test suite for wombat using [ava](https://github.com/avajs/ava) (I use it in [WAIL](https://github.com/N0taN3rd/wail/blob/wail-electron/tests/testWail.js)) and [electron](https://electron.atom.io/) or continuing
documentation etc.

N0taN3rd:
@ikreymer any life for this PR? 
Rebased and got  ``N0taN3rd:wombatProxies`` up to  ``ikreymer:new-pywb`` or at least it seems that way on my end.

ikreymer:
Closing, this is now been implemented via #224

PR #258, opened 2017-10-19
New Query.html

Fernando-Melo:
New query.html page responsive made with bootstrap + jquery


ikreymer:
Thanks! Closing, just replacing with #272 which adds some additional cleanup

PR #279, opened 2018-01-16
make manager commands working with python3

atomotic:
python3: `raw_input()` was renamed to `input()`


ikreymer:
Thanks! missed it as input was being mocked up in tests

PR #308, opened 2018-03-05
rewrite_amf.py: Fix bug introduced by recent refactoring

humberthardy:
Small bug introduced by commit #d8b6731

### DIFF
--- a/pywb/rewrite/rewrite_amf.py
+++ b/pywb/rewrite/rewrite_amf.py
@@ -22,7 +22,7 @@ class RewriteAMF(BufferedRewriter):  #pragma: no cover
             res = remoting.decode(iobuff)
 
             # TODO: revisit this
-            inputdata = url_rewriter.rewrite_opts.get('pywb.inputdata')
+            inputdata = rwinfo.url_rewriter.rewrite_opts.get('pywb.inputdata')
 
             if inputdata:
                 new_list = []


ikreymer:
@humberthardy thanks for the fix.. this feature is still very experimental, it was implemented for a specific example of AMF, but haven't had a chance to look at this further.

Do you have a test case where AMF if used? If so, would you mind sending it along?
Or did the error pop up in some other way? Just curious

humberthardy:
Hello @ikreymer 
I have captured one of our dead sites (with the help of warcprox + docker to try to make it alive locally).
Then, I have tried to replay it with webrecorder when I have found this problem. With the fix, the first request works but I got stuck just after. After having a deeper look, It seems that https://github.com/webrecorder/pywb/blob/master/pywb/warcserver/inputrequest.py#L265 is looking for some magic values inside the AMF, but in our website's webservices are not in the same format, so (I guess) responses are not related to the right RPC call.


Here is our format. The `target` attribute is like your `operation` one and the parameters are the full body. (python's representation after pyamf's parsing):

```
<Envelope amfVersion=0>
    ('/1', <Request target='.atoms.load'>[{'slug': 'ONF828595'}]</Request>)
</Envelope>
```

I have tried to fix it with more generic function (try to store the python's representation in the `query` variable), but without any success for now. To be honest, I am not sure to fully understand how indexation & url matching is working with pyweb & webRecorder.

If you are interested to see more, you can find an example here: https://webrecorder.io/hhardy/amf/20180302183425$br:chrome:53/http://pib.onf.ca/episode/1731/au-revoir-yale
The responses from the proxies are not correct and the flash's player disappear.

If you would like more information or more help, feel free to ask!

Humbert.

humberthardy:
Hi  @ikreymer

I think I get it,
for some reasons, in this example, the representation of the same object can be different, because, if I get it, serialization of objects does not preserve the order of attributes.
For example, here the representation of the `AMF` request inside the `warc` file.

```
<Request target='.reactions.search'>[{'searchOptions': {'relations': ['user', 'documents']}, 'searchParams': {'type': ['answer_video', 'answer_image', 'answer_text'], 'status': 200, 'lang': ['all', 'FR'], 'atom_id': 1809}}]</Request>
```

And, here, we have a request made inside the browser: It is the same object but not the same representation as (see lang parameter is not in the same order)

```
<Request target='.reactions.search'>[{'searchOptions': {'relations': ['user', 'documents']}, 'searchParams': {'lang': ['all', 'FR'], 'type': ['answer_video', 'answer_image', 'answer_text'], 'status': 200, 'atom_id': 1809}}]</Request>
```

I have written a quick fix that seems to work well (for this project at least). 

https://github.com/webrecorder/pywb/commit/4ddcd27bdc00473284ed16e8a5d0947e9e3c8495

I am a little bit shameful about it because it is more like a hack, it computes a hash from letters and words inside the representation of the request. 
Feel free to delete the branch `improve-amf-request` if you want.

I do not have open a pull request yet, I am waiting for your feedback first. If you have a better idea, do not hesitate to share. I know that rewrite the representation of `AMF` object can fix the problem in a better way, but it seems to be a huge work.

I know that it is better than it was, but it is far to be perfect: If I remember well, in AMF, you can have multiple requests in one `HTTP` request. When you replay the project, `Flash player` can choose to spit/merge requests to one/many `http` requests. But I do not have met this problem yet.

Humbert.

PR #320, opened 2018-05-16
Improve indexing for amf requests

humberthardy:
The current version of pywb have some problems with AMF:
- It was developed to only support AMF messages from Adobe Flex (Adobe Flex is a Flash framework)
- The body of a request can be JSON, and the order of attribute can be a problem.
 
Here, we have developed a small class that gives a string representation of an Amf request. It supports pure AMF and Flex messages and tries to reorder the attributes of JSON objects.

Here are some pure AMF websites:

1.  [**matribu**](http://matribu.onf.ca/#/matribu)  -  [replay link](https://webrecorder.io/hhardy/amf/20180420140548$br:chrome:53/http://matribu.onf.ca/#/matribu) : With the patch, you will be able to record and replay the project.   
2.  **PIB** -  [replay link](https://webrecorder.io/hhardy/amf/http://pib.onf.ca/)To see movies, you need RTMP server (not HTTP), but, at least, you should be able to browse the website with the patch.

Unfortunately, I do not have any Flex projects to test support of it. (only the unit test) If you have one, I will be glad to test it and give you the feedback. 

Humbert.

### DIFF
--- a/extra_requirements.txt
+++ b/extra_requirements.txt
@@ -2,6 +2,5 @@ certauth
 youtube-dl
 boto3
 uwsgi
-git+https://github.com/t0m/pyamf.git@python3
 git+https://github.com/esnme/ultrajson.git
 pysocks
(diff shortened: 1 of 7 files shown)


humberthardy:
will fix python 2.7 problems before reopening it

PR #321, opened 2018-05-17
Handle amf requests

humberthardy:
The current version of `pywb` have some problems with `AMF`:
- It was developed to only support `AMF` messages from `Adobe Flex` (`Adobe Flex` is a Flash framework)
- The body of a request can be JSON, and the order of attribute can be a problem.
 
Here, we have developed a small class that gives a string representation of an `AMF` request. It supports pure `AMF` and `Flex` messages and tries to reorder the attributes of JSON objects.

## Examples
Here are some pure `AMF` websites:

1.  [**matribu**](http://matribu.onf.ca/#/matribu)  -  [replay link](https://webrecorder.io/hhardy/amf/20180420140548$br:chrome:53/http://matribu.onf.ca/#/matribu) : With the patch, you will be able to record and replay the project.   
2.  **PIB** -  [replay link](https://webrecorder.io/hhardy/amf/http://pib.onf.ca/)To see movies, you need RTMP server (not HTTP), but, at least, you should be able to browse the website with the patch.

Unfortunately, I do not have any Flex projects to test support of it. (only the unit test has been done for this part) If you have one, I will be glad to test it and give you the feedback. 

## A note about python compatibility
The only amf library for python supporting 2.7,  3 and remoting messages is `git+https://github.com/t0m/pyamf.git@python3` (not available in `PyPI`).
 
I have modified `setup.py` to install either `pyamf` or `py3amf` depending on the version of python.

### DIFF
--- a/extra_requirements.txt
+++ b/extra_requirements.txt
@@ -2,6 +2,5 @@ certauth
 youtube-dl
 boto3
 uwsgi
-git+https://github.com/t0m/pyamf.git@python3
 git+https://github.com/esnme/ultrajson.git
 pysocks
--- a/pywb/rewrite/rewrite_amf.py
+++ b/pywb/rewrite/rewrite_amf.py
@@ -43,6 +43,8 @@ class RewriteAMF(BufferedRewriter):  #pragma: no cover
             import traceback
             traceback.print_exc()
             print(e)
+
+            stream.seek(0)
             return stream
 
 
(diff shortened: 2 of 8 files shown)


ikreymer:
Thanks very much for this PR!
I was able to test "matribu", but not "pib", as it appears to be offline at the moment.
The only Flex example I had from before did not work, but it also wasn't working with an earlier version either. Perhaps its not worth spending more time on it until there is a more concrete use case.

ikreymer:
My only other concern was making the pyamf libraries required (before, pyamf was only installed via extra_requirements.txt), but seems that they install well for both py2 and py3, so let's give it a go. If there are issues, can make them optional again.
Thanks again for all your work on this.

PR #323, opened 2018-05-17
serviceworker postMessage rewrite fix

### LINKED ISSUE #322: Lack of communication with serviceworkers
## Expected Behavior
A page should be able to communicate with its registered service worker via [postMessage](https://w3c.github.io/ServiceWorker/#serviceworker-interface) and or a [Message Channel](https://html.spec.whatwg.org/multipage/web-messaging.html#message-channels).

Likewise it is expected that the registered service worker can communicate with its controlled clients via [postMessage](https://w3c.github.io/ServiceWorker/#client-postmessage) or a [message port](https://html.spec.whatwg.org/multipage/web-messaging.html#messageport) on  a [message event](https://w3c.github.io/ServiceWorker/#extendablemessageevent-interface).

For example consider the following live web page: https://n0tan3rd.github.io/replay_test 
It registers a service worker and attempts to communicate with it via postMessage and a message channel
- [page js src](https://github.com/N0taN3rd/replay_test/blob/master/src/epics/swEpics.js)
- [service worker src](https://github.com/N0taN3rd/replay_test/blob/master/replayTest_SW.js)

## Actual Behavior
https://webrecorder.io/jberlin/sw/20180510171123/https://n0tan3rd.github.io/replay_test/
serviceworkerbadrw

The current strategy employed by Pywb does supply postMessage rewrite fix for service workers 
https://github.com/webrecorder/pywb/blob/9acad278011e7ec6e2b4e2044e137d5d5382c840/pywb/rewrite/regex_rewriters.py#L160-L164

## Steps to Reproduce the Problem
Replay https://webrecorder.io/jberlin/sw/20180510171123/https://n0tan3rd.github.io/replay_test/

## Context
- Pywb Version:  2.0.3

N0taN3rd:
This PR is to fix #322 by patching the prototypes of [Client](https://w3c.github.io/ServiceWorker/#client-interface) and [MessagePort](https://html.spec.whatwg.org/multipage/web-messaging.html#message-ports) interfaces if the client interface is not undefined.

To understand how this works consider the following IDLs:

```webidl
[Exposed=ServiceWorker]
interface Client {
  readonly attribute USVString url;
  readonly attribute DOMString id;
  readonly attribute ClientType type;
  void postMessage(any message, optional sequence<object> transfer = []);
};
```
The client interface is only exposed on the execution environments global object for service workers.
We can use this information to guard the override of ```Client.postMessage```.
```js
if (!self.__WB_pmw && typeof self.Client !== "undefined") 
```
The patch override applied to ```Client.postMessage```  by this PR
```js
self.Client.prototype.__WB_pmw = function(obj) { return this; }
```
differs from the current convention for handling non existing ```__WB_pmw``` properties
```js
self.__WB_pmw = function(obj) { return obj; }
```
because the message needs to be sent directly to the controlled client.

By returning ```this``` rather than the ```obj``` param of the function, we ensure that the rewritten ```postMessage``` call
```js
client.__WB_pmw(self).postMessage(message);
```
uses the client interface not the service worker.

Likewise, the patch override of ```MessagePort.postMessage``` follows the same reasoning.
```js
// patch override
self.MessagePort.prototype.__WB_pmw = function(obj) { return this; };

// rewritten
event.ports[0].__WB_pmw(self).postMessage(event.data.m + " SW Says 'Hello back!'");
```
In the context of a service worker the message port is found on an [```ExtendableMessageEvent ```](https://w3c.github.io/ServiceWorker/#extendablemessageevent-interface), a  [```ServiceWorkerMessageEvent```](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerMessageEvent).
```webidl
[Constructor(DOMString type, optional ExtendableMessageEventInit eventInitDict), Exposed=ServiceWorker]
interface ExtendableMessageEvent : ExtendableEvent {
  readonly attribute any data;
  readonly attribute USVString origin;
  readonly attribute DOMString lastEventId;
  [SameObject] readonly attribute (Client or ServiceWorker or MessagePort)? source;
  readonly attribute FrozenArray<MessagePort> ports;
};

[Exposed=(Window,Worker,AudioWorklet), Transferable]
interface MessagePort : EventTarget {
  void postMessage(any message, optional sequence<object> transfer = []);
  void start();
  void close();

  // event handlers
  attribute EventHandler onmessage;
  attribute EventHandler onmessageerror;
};
```

### DIFF
--- a/pywb/rewrite/regex_rewriters.py
+++ b/pywb/rewrite/regex_rewriters.py
@@ -160,6 +160,10 @@ class JSWombatProxyRewriterMixin(object):
     local_init_func = '\nvar {0} = function(name) {{\
 return (self._wb_wombat && self._wb_wombat.local_init &&\
  self._wb_wombat.local_init(name)) || self[name]; }};\n\
+if (!self.__WB_pmw && typeof self.Client !== "undefined") {{ \n\
+  self.Client.prototype.__WB_pmw = function(obj) {{ return this; }}; \n\
+  self.MessagePort.prototype.__WB_pmw = function(obj) {{ return this; }}; \n\
+}}\n\
 if (!self.__WB_pmw) {{ self.__WB_pmw = function(obj) {{ return obj; }} }}\n\
 {{\n'
 


N0taN3rd:
Ping @ikreymer for review

ikreymer:
Summarizing from discussion over Slack. The fix works well for the case service worker used on: https://n0tan3rd.github.io/replay_test but actually identified two other issues!

1) The `init_server_worker_override()` in wombat.js is supposed to be setting the modifier to `id_` which should prevent any rewriting for being applied to service workers. However, since only a filename is used, the link is not being rewritten. If it had worked as expected, this fix actually would not have applied to the service worker on https://n0tan3rd.github.io/replay_test

2) Ideally, there would be a separate modifier just for service workers, so that standard document insertion JS would not needed to be applied. Perhaps `sw_`. This modifier could then be used to rewrite scripts that are used as service workers, and this change could be added only for those scripts. This would mean adding a new modifier and is more flexible in the long run.

N0taN3rd:
I agree that adding a specific modifier for service workers (`sw_`) is the most flexible option in the long run. 
Without the new modifier, detecting that we are rewriting service worker specific JS server-side would not be feasible due to the amount of analysis required to make the determination.
The `importScripts` function is not a unique enough identifier given it is used in both the service worker and web worker contexts, as well as, it is a viable identifier in non-worker contexts.

ikreymer:
Closing, issues raised here fixed by #339 instead for now.

PR #325, opened 2018-05-18
Fix wombats overrides of document.[write, writeln] to account for the variadic case

### LINKED ISSUE #324: Current wombat overrides for document.[write,writeln] do not handle the variadic case
The current overrides of document.[write, writeln] do not handle the case when there are multiple arguments, [spec](https://html.spec.whatwg.org/multipage/dom.html#the-document-object).
This affects replace of pages that rely on this behavior.
- [page](http://www.easter.ne.jp)
- [script using document.write with multiple args](http://www.textad.net:10001/cgi-bin/manager.cgi?category_id=0&i=1)
https://github.com/webrecorder/pywb/blob/33cca0bc026acc2da1cdace8c701776900118166/pywb/static/wombat.js#L2613-L2621

https://github.com/webrecorder/pywb/blob/33cca0bc026acc2da1cdace8c701776900118166/pywb/static/wombat.js#L2629-L2637

N0taN3rd:
This PR fixes #324 by check for the variadic case in the document.[write, writeln] function overrides.   

For reference consider the [Document IDL](https://html.spec.whatwg.org/multipage/dom.html#the-document-object)  excerpt below
```webidl
[OverrideBuiltins]
partial interface Document {
  [CEReactions] void write(DOMString... text);
  [CEReactions] void writeln(DOMString... text);
}
```
The changes made by this PR can be summarized as:
1. Added a reference to `$wbwindow.Function.prototype.apply` called `origFunctionApply` to account for the Function.apply override
2. Made the argument string (from the original override) a function variable 
3. If the length of the `arguments` object is less than or equal to 1, set string to `arguments[0]`
4. Otherwise convert the `arguments` object to an array via `origFunctionApply.call($wbwindow.Array, arguments);` and set string variable to the results of `argArray.join('')`

### DIFF
--- a/pywb/static/wombat.js
+++ b/pywb/static/wombat.js
@@ -2647,7 +2647,7 @@ var _WBWombat = function($wbwindow, wbinfo) {
             var argLen = arguments.length;
             var string;
             if (argLen === 0) {
-                return orig_doc_write.call(this);
+                return orig_doc_writeln.call(this);
             } else if (argLen === 1) {
                 string = arguments[0];
             } else {
--- a/pywb/static/wombat.js
+++ b/pywb/static/wombat.js
@@ -92,6 +92,9 @@ var _WBWombat = function($wbwindow, wbinfo) {
     var message_listeners = new FuncMap();
     var storage_listeners = new FuncMap();
 
+    // to avoid incurring the cost of our override
+    var origFunctionApply = $wbwindow.Function.prototype.apply;
+
     //============================================
     function is_host_url(str) {
         // Good guess that's its a hostname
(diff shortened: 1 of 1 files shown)


pywb/static/wombat.js
```diff
@@ -2607,34 +2610,52 @@ var _WBWombat = function($wbwindow, wbinfo) {
...
+
         // Write
         var orig_doc_write = $wbwindow.document.write;
 
-        var new_write = function(string) {
-            new_buff = rewrite_html(string, true);
+        var new_write = function() {
+            var string;
+            if (arguments.length <= 1) {
+                string = arguments[0];
```
ikreymer: Should this have a separate check for `arguments.length == 0` as `document.write()` is technically allowed? Maybe just call `orig_doc_write.call(this)` and exit?

pywb/static/wombat.js
```diff
@@ -2607,34 +2610,52 @@ var _WBWombat = function($wbwindow, wbinfo) {
...
         $wbwindow.Document.prototype.write = new_write;
 
         // Writeln
         var orig_doc_writeln = $wbwindow.document.writeln;
 
-        var new_writeln = function(string) {
-            new_buff = rewrite_html(string, true);
+        var new_writeln = function() {
+            var string;
+            if (arguments.length <= 1) {
```
ikreymer: Same as above re: zero arguments case.

pywb/static/wombat.js
N0taN3rd: You are correct, `arguments.length == 0` as `document.write() ` is allowed. 
Added this case.

pywb/static/wombat.js
N0taN3rd: Case added

PR #334, opened 2018-05-29
Server-side rewriting of script[src='js/...'] and link rel='import'

N0taN3rd:
### Updated html_rewriter.py to account for rewriting of script[src] values that are super relative 
URL requiring change: http://fotopaulmartens.netcam.nl/vucht.php
Script tags with src='x' where x is not absolute, relative, schemeless, or contain `../` resolve (i.e. `js/scplayer.php?camera=paulmartens102129&server=6&port=443&protocol=https`) are resolved by the browser to include a non-js rewrite modifier.
While this is fine in most cases there exists cases where the server does not always send the correct JS MIME type causing JS to not be rewritten as JS. 
In this case PYWB will now turn the "super relative" URL into an absolute URL with the correct rewrite modifier (`js_`) and add an attribute, `__wb_orig_src` to the element containing the unrerwritten URL for wombat.

This case is detected by `html_rewriter.py` using the new method `_not_abs_rel_has_parentp` which contains some duplicated logic from `url_rewriter.py` in order to leave `url_rewriter.py` un-modified.
The check made by `_not_abs_rel_has_parentp` is the opposite of the condition that allowed this to happen before this PR with special consideration to ensure we are not rewriting values not rewritten ever.

### Updated wombat to account for the new rewriting of script[src] 
URL requiring change: http://fotopaulmartens.netcam.nl/vucht.php

Wombat was modified in order to operate correctly under this change when `getAttribute` is used with special consideration to browser resolution of `src, href` attributes when being set.
When the script tag with the `__wb_orig_src` has its `src` attribute changed the `__wb_orig_src` attribute is removed and an internal flag `__$removedWBOSRC$__`, existing on the element itself, is set.
This is done so that further checks for the  `getAttribute` case do not have to made.
In `getAttribute` the value of `__wb_orig_src`  is only returned for script tags if this attribute exists and the  `__$removedWBOSRC$__` flag is false  or undefined.
Removal of  `__wb_orig_src` is handled by `handleWbOrigSrc`.

A minimal working example of this can be found at https://codepen.io/anon/pen/RJbqye.

### Updated html_rewriter.py  link rel='import' rewriting
URL requiring change: https://www.youtube.com
Only change was as check in `_rewrite_link_href` for `rel='import'` and if so `rw_mod = 'mp_'`

### Updated test_html_rewriter.py for super rel script[src] rewriting and link rel='import'
Reasoning: see above

### DIFF
--- a/pywb/static/wombat.js
+++ b/pywb/static/wombat.js
@@ -1089,6 +1089,7 @@ var _WBWombat = function($wbwindow, wbinfo) {
     }
 
     //============================================
+
     function init_setAttribute_override()
     {
         if (!$wbwindow.Element ||
@@ -1110,6 +1111,12 @@ var _WBWombat = function($wbwindow, wbinfo) {
                     value = rewrite_inline_style(value);
 
                 } else if (should_rewrite_attr(this.tagName, lowername)) {
+                    if (this.tagName === 'SCRIPT' && !this.__$removedWBOSRC$__) {
+                        if (this.hasAttribute("__wb_orig_src")) {
+                            this.removeAttribute("__wb_orig_src");
+                        }
+                        this.__$removedWBOSRC$__ = true;
+                    }
                     if (!this._no_rewrite) {
                         var mod = rwModForElement(this, lowername);
                         value = rewrite_url(value, false, mod);
@@ -1129,7 +1136,7 @@ var _WBWombat = function($wbwindow, wbinfo) {
     {
         if (!$wbwindow.Element ||
             !$wbwindow.Element.prototype ||
-            !$wbwindow.Element.prototype.setAttribute) {
+            !$wbwindow.Element.prototype.getAttribute) {
             return;
         }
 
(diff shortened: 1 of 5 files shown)


N0taN3rd:
The `postMessage` fixes that were originally apart of this PR was split into its own PR https://github.com/webrecorder/pywb/pull/338 per Slack discussion

pywb/static/wombat.js
```diff
@@ -1110,6 +1111,12 @@ var _WBWombat = function($wbwindow, wbinfo) {
                     value = rewrite_inline_style(value);
 
                 } else if (should_rewrite_attr(this.tagName, lowername)) {
+                    if (this.tagName === 'SCRIPT' && !this.__$removedWBOSRC$__) {
```
ikreymer: Can this logic be moved to its own function, and then just have a` _check_wbosrc(elem)` since its repeated 3 times? Will make maintenance easier. And then should be good!

N0taN3rd: Done

PR #339, opened 2018-05-29
ServiceWorker Rewrite Improvements

ikreymer:
Experimental SW rewriting fixes, attempts to address issues raised in #323:
- ServiceWorker override resolved to absolute url before rewriting to use the `sw_` modifier.
- `sw_` modifier is similar to `id_`, but adds `Service-Worker-Allowed` to domain root
- wombat: service worker override rewrites origin, setting a default origin to domain root with mp_ modifier (to match Service-Worker-Allowed header)
- wombat init: don't call wombat init_paths if wombat already inited, as may happen from imported document.


ikreymer:
WIP - still need to add test, but should work on: https://n0tan3rd.github.io/replay_test/

PR #347, opened 2018-06-14
Gevent requirement update 

N0taN3rd:
This PR is to update the gevent requirement from 1.2.2 to 1.3.3.
The changes made by this PR relatively small but come with performance increases.

### Gevent 1.3.3
Gevent, since 1.3, now using Cython for performance critical portions of the code base.
However, with the update there are planned deprecations and new perf recommendations.

The old c-ares DNS resolver used by gevent is considered "deprecated" (not officially) and in the next major revision will be removed. 
Gevent recommends the usage of `dnspython` because of its performance increases. 
Also idna was explicitly added  as it is for py2.

### pywb code changes
The only pywb core code changes made by this PR is change the `gevent.wsgi` import used by `geventserver.py` to `gevent.pywsgi`.

`gevent.wsgi` was an alias for `gevent.pywsgi` and has been removed.


requirements.txt
```diff
@@ -9,7 +9,9 @@ brotlipy
 pyyaml
 werkzeug
 webencodings
-gevent==1.2.2
+dnspython     # for updated gevent: recommend by gevent, c-ares resolver to be deprecated soon
+idna          # for updated gevent: Python 2.7 support
+gevent==1.3.3 # update requirement. Linux needs manual source install
```
ikreymer: Seems like 1.3.4 for is out and has some wsgi related optimizations (http://www.gevent.org/changelog.html#id1) so should upgrade to it instead.

requirements.txt
```diff
@@ -9,7 +9,9 @@ brotlipy
 pyyaml
 werkzeug
 webencodings
-gevent==1.2.2
+dnspython     # for updated gevent: recommend by gevent, c-ares resolver to be deprecated soon
+idna          # for updated gevent: Python 2.7 support
```
ikreymer: It looks like `idna` is already required by `requests` so probably don't need to include it explicitly

N0taN3rd: done

requirements.txt
N0taN3rd: done

PR #349, opened 2018-06-20
Facebook and Web Worker fixes

N0taN3rd:
URLs requiring change: https://www.facebook.com/ and https://spotify---bastille.appspot.com/
#### Wombat changes
Facebook polyfills requestAnimationFrame and when we return the polyfill function bound to window they can detect that and complain. 
The fix is to not return a bound function when `prop` is `requestAnimationFrame` **IFF** Function.toString does not return `"native code"`
- default_proxy_get improvement: Facebook fix: if prop is requestAnimationFrame (raf) and it was polyfilled by FB do not bind

Both Facebook, the spotify app and many other websites use `Web Workers (Worker)` that are created not using blobs; likewise with `ShardWorker`. 
The previous override only handled the blob case for `Worker` and thus any other use case was not handled properly.
The fix is to expand on the rewriting of worker urls for both `Worker` and `SharedWorker`. 
Includes a new override for `SharedWorker` and introduces a new modifier `wkr_`.
- improved worker rewriting: updated worker rewriting handles non-blob urls, added SharedWorker override

#### ww_rw.js changes
The min-rewriting system for workers was updated to provide a more complete system that now handles 
`importScripts`, `Worker.fetch`, and an `eval` edge case unique to the spotify app.   
- updated to be a much more complete rewriting system: overrides for importScripts, fetch, and eval "hack"

#### content_rewriter.py changes
In order to ensure the setup for the full client-side rewriting system is not included in worker code the `wkr_` modifier was introduced.
This modifier behaves exactly like `sw_`.
 - added `wkr_` mod for handling Worker/SharedWorker, follows convention of service worker

#### test_content_rewriter.py changes
The tests for `content_rewriter.py` was updated for the introduction of the new modifer `wkr_`.
- added test for content rewriting of Worker/SharedWorker

### DIFF
--- a/pywb/rewrite/content_rewriter.py
+++ b/pywb/rewrite/content_rewriter.py
@@ -381,7 +381,7 @@ class RewriteInfo(object):
     def _resolve_text_type(self, text_type):
         mod = self.url_rewriter.wburl.mod
 
-        if mod == 'sw_':
+        if mod == 'sw_' or mod == 'wkr_':
             return None
 
         if text_type == 'css' and mod == 'js_':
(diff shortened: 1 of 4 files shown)


N0taN3rd:
closed via PRs #350 and #351

PR #354, opened 2018-06-28
RegexRewriter Optimization

ikreymer:
- Split RxRewriter rules into separate class hierarchy, inited once
- Each RegexRewriter has a 'rules factory' which generates the rules for the rewriter
- If extra rules provided, recompile with dynamic rules, otherwise return existing fixed rules
- Avoids recompiling rules rx for each rewriter (unless dynamic rules used), instead store rules object per rewriter type

### DIFF
--- a/pywb/rewrite/regex_rewriters.py
+++ b/pywb/rewrite/regex_rewriters.py
@@ -5,10 +5,12 @@ from six.moves.urllib.parse import unquote
 
 
 # =================================================================
-class RegexRewriter(StreamingRewriter):
-    # @staticmethod
-    # def comment_out(string):
-    #    return '/*' + string + '*/'
+class RxRules(object):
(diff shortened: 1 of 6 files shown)


pywb/rewrite/regex_rewriters.py
```diff
@@ -231,34 +270,12 @@ def final_read(self):
...
-# =================================================================
-class JSLinkAndLocationRewriter(JSLocationRewriterMixin,
-                                JSLinkRewriterMixin,
-                                RegexRewriter):
-    pass
-
-
 # =================================================================
 class JSNoneRewriter(RegexRewriter):
     def __init__(self, rewriter, rules=[]):
```
N0taN3rd: I think it maybe wise to change `rules=[]` to
```python
def __init__(self, rewriter, rules=None):
    if rules is None:
        rules = []
``` 
The default assignment to `[]` is a perf hit as the initial assignment persists across all creations and you never get a "new" list again.

PR #359, opened 2018-07-24
Advanced Srcset and Style Sheet Containing Media Queries Preservation

N0taN3rd:
## Description

This PR introduces to Pywb via changes to `wombat.js` and the addition of a new static resource `wombatPreservationWorker.js` the ability to preserve 
the set of values contained in the `srcset` attribute and images contained in `CSS media query rules`.

The preservation scheme is split into two parts:
1. `wombat`: Initiating `wombatPreservationWorker` and sending it `srcset` and `media query` values for preservation
2. `wombatPreservationWorker`: The [web worker](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API) responsible for fetching the resources

### wombat
In order to facilitate this new preservation scheme the changes to wombat are as follows. 

##### Three new internal global variables
1. `WBPreserWorker`: a reference to the preservation worker abstraction used by wombat itself
2. `wbSheetMediaQChecker`: a reference to the `load` event callback for checking the stylesheet introduced by `link[rel='stylesheet']`
3. `wbUsePresWorker`: flag indicating if we should use the preservation worker. 

The first two global variables (`WBPreserWorker` and `wbSheetMediaQChecker`) remain undefined unless the third (`wbUsePresWorker`) is true.
`wbUsePresWorker` is set to true when `window.Worker` is non-null and wombat is operating in live mode or the `wbinfo.top_url` contains `record`.

##### New init function initPreserveWorker and PWorker
The new init function, `initPreserveWorker`, contains wombats interface to the new `wombatPreservationWorker`, that was designed to be cross-frame safe.
The init function only introduces the `PWorker` interface when the `wbUsePresWorker` flag is true.

`initPreserveWorker` operates similar to the existing `web worker` override except that it must come first in order to use non-overridden reference to `window.Worker`
which is used by `PWorker` (preservation worker).

`PWorker` is an ES5 class that provides functionality for cross-frame preservation of `srcset` and `media query` CSS rule preservation. 
Creation of the `PWorker` takes two arguments `prefix` (`wb_abs_prefix`) and `mod` (`wbinfo.mod`).
If wombat is operating in the `__WB_replay_top` browser context, the backing worker is created otherwise it is not re-created.
The URL to the backing worker includes query parameters `prefix` and `mod` which are used by `wombatPreservationWorker` to create itself if we can use `window.URL`
otherwise we must immediately send an `init` message containing those values to the worker.  

The `PWorker` class exposes six functions:
1. `deferredSheetExtraction`: Method for checking a `style` or `link[rel='stylesheet']` elements `sheet` property for media query rules. 
Done in a manner that does not block the UI thread by using `Promise.resolve` (when available) or `setTimeout(cb, 1)`.
Used by the `checkStyle` (value of `wbSheetMediaQChecker`), `rewrite_elem` and `override_html_assign` functions.
Usage in `rewrite_elem` is tied to elements `style` (text contents changed and `sheet != null`) and `link[rel='stylesheet']` (via `link.addEventListener('load',wbSheetMediaQChecker)`).

2. `terminate`: terminate the backing web worker, only terminates worker when in the `__WB_replay_top` browser context.

3.  `postMessage`: send a message to the backing web worker. If wombat is in the `__WB_replay_top` browser context sends message directly 
to worker otherwise forwards msg to `__WB_replay_top` which will then send the msg directly to the backing worker.

4. `preserveSrcset`: method used when sending `srcset` values from `rewrite_srcset`
5. `preserveMedia`: method used when sending `media query` values from `deferredSheetExtraction` 
6. `extractFromLocalDoc`: method to check `document.stylesheets` and `document.querySelectorAll('*[srcset]')` when `notify_top` is called from `init_top_frame_notify`

### wombatPreservationWorker
The new file `wombatPreservationWorker.js` contains the worker code used for preserving `srcset` and `media query` values.
Because it is expected of wombat to operate in a wide range of browsers this file contains minimal polyfills of `Promise` and `fetch` if these symbols are not defined.

This worker expects to receive two messages from wombat: 
1. `init`: when the [WHATWG URL](https://url.spec.whatwg.org/#dom-url) parser is not available to instantiate the Preserver class.
2. `values`: contains values to be preserved.
    
The `Preserver` class (ES5) is responsible for the extraction and preservation of the desired values.
The `Preserver` tracks seen URLs, up to 2500 before resetting, in order to lessen the load against the server(s) requests are made too.
If the `Preserver` has seen a URL no fetch is made otherwise it is added to the seen cache.

The `Preserver` class exposes 9 functions:
1. `fixupURL`: Ensures the URL to be fetched are absolute 
2. `safeFetch`: Initiates a URL fetch or queues the URL to be fetched if we are waiting for a batch of fetches to complete
3. `urlExtractor`: Extracts the URL from sheet text (media query) in a similar manner as `style_replacer`
4. `fetchDone`: When a batch of fetches is done ensures the URL queue is drained and if the `Preserver` has seen 2500 URLs clears the seen count for GC purposes.
5. `fetchAll`: Ensures all fetches complete, `fetchDone` is called and indicates we are to queue URLs until fetches complete. 
If `Preserver` is queuing URLs or no fetches are to be made this is a no op.
6. `drainQ`: Drains the queue by calling `safeFetch` for each queued URL and then calls `fetchAll`
7. `extractMedia`: For each style sheet that contained a media query, extract URLs from it's text and fetch initiate a fetch for it if there was a URL 
8. `extractSrcset`: For each `srcset` value, if the value is from `Pworker.preserveSrcset` only fetch URL otherwise from `Pworker.extractFromLocalDoc` need to extract split full `srcset` attribute value and then fetch each URL contained
9. `preserveMediaSrcset`: Used when the `values` message is received and calls `extractMedia`, `extractSrcset` and `fetchAll`.

## Motivation and Context

Ref https://github.com/webrecorder/webrecorder/issues/64

## Types of changes

### DIFF
--- a/pywb/static/wombat.js
+++ b/pywb/static/wombat.js
@@ -78,6 +78,9 @@ var _WBWombat = function($wbwindow, wbinfo) {
     var wb_setAttribute = $wbwindow.Element.prototype.setAttribute;
     var wb_getAttribute = $wbwindow.Element.prototype.getAttribute;
     var wb_funToString = Function.prototype.toString;
+    var WBPreserWorker;
+    var wbSheetMediaQChecker;
+    var wbUsePresWorker = $wbwindow.Worker != null && wbinfo.is_live;
 
     var wb_info;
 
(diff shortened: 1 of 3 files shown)


pywb/static/wombat.js
```diff
@@ -1326,6 +1330,144 @@ var _WBWombat = function($wbwindow, wbinfo) {
...
+                        }
+                    }
+                    if (media.length > 0) {
+                        // we have some media rules to preserve
+                        WBPreserWorker.preserveMedia(media);
+                    }
+                }
+                // depending on if we have promises, defer things until
+                // next time the Promise.resolve or setTimeout Qs are cleared
+                if (typeof $wbwindow.Promise !== 'undefined') {
```
ikreymer: Is it safe to assume we have Promises if we have workers? If so, maybe should just do it one or the other, as there's already enough conditionals here. :) I think we use Promises already in wombat..

pywb/static/wombat.js
```diff
@@ -78,6 +78,10 @@ var _WBWombat = function($wbwindow, wbinfo) {
     var wb_setAttribute = $wbwindow.Element.prototype.setAttribute;
     var wb_getAttribute = $wbwindow.Element.prototype.getAttribute;
     var wb_funToString = Function.prototype.toString;
+    var WBPreserWorker;
+    var wbSheetMediaQChecker;
+    var wbUsePresWorker = $wbwindow.Worker != null &&
+        (wbinfo.is_live || wbinfo.top_url.indexOf('/record/') !== -1);
```
ikreymer: I think we shouldn't need the `/record/` check, `is_live` should be sufficient..

pywb/static/wombat.js
```diff
@@ -1517,11 +1659,13 @@ var _WBWombat = function($wbwindow, wbinfo) {
 
         // Filter removes non-truthy values like null, undefined, and ""
         var values = value.split(/\s*(\S*\s+[\d\.]+[wx]),|(?:\s*,(?:\s+|(?=https?:)))/).filter(Boolean);
-
         for (var i = 0; i < values.length; i++) {
             values[i] = rewrite_url(values[i].trim());
         }
-
+        if (WBPreserWorker) {
```
ikreymer: Should this be checking `wbUsePresWorker` also?

pywb/static/wombat.js
```diff
@@ -3630,6 +3804,9 @@ var _WBWombat = function($wbwindow, wbinfo) {
             }
 
             send_top_message(message);
+            if (WBPreserWorker) {
```
ikreymer: check `wbUsePresWorker` for consistency?

pywb/static/wombat.js
```diff
@@ -1326,6 +1330,144 @@ var _WBWombat = function($wbwindow, wbinfo) {
...
+                }
+                if ($wbwindow === $wbwindow.__WB_replay_top) {
+                    // we are top and can will own this worker
+                    // setup URL for the kewl case
+                    var workerURL = wbinfo.static_prefix +
+                        'wombatPreservationWorker.js?prefix=' +
+                        encodeURIComponent(prefix) + '&mod=' +
+                        encodeURIComponent(mod);
+                    this.worker = new Worker(workerURL);
+                    if (typeof $wbwindow.URL === 'undefined') {
```
ikreymer: Maybe this should just be the default, to avoid the multiple init paths? (I wonder which browsers support workers and not window.URL..) Just trying to remove variation :)

pywb/static/wombat.js
ikreymer: `notify_top()` actually runs twice (for readyState `interactive` and for readyState `complete`, to allow the banner to update as soon as possible). Probably should send to worker only when readyState is `complete`?

pywb/static/wombat.js
N0taN3rd: Right you are. 
`is_live` is sufficient, changing

pywb/static/wombat.js
N0taN3rd: It would seem that safari is the only browser that might give us trouble is safari (`searchParams` available in v10.1)

N0taN3rd:
Probably best to add proxy support in another task because we would have to check periodically and thus be sending repeat values to the worker depending on the amount of time the browser stays on the same page.

A more involved deduplication strategy is in order to account for this fact

PR #360, opened 2018-07-26
Enhance help messages in the command line interface

marhop:
## Description
Added help messages to the `add_argument` method calls in class `BaseCli` where none were present yet. This documents the meaning and default values of the available CLI options.

## Motivation and Context
It took me some time to find out e.g. the meaning and default value of the `--bind` option. Now a user won't have to search the docs or the source code for that.

## Types of changes

## Checklist:

### DIFF
--- a/pywb/apps/cli.py
+++ b/pywb/apps/cli.py
@@ -29,17 +29,24 @@ def live_rewrite_server(args=None):
 class BaseCli(object):
     def __init__(self, args=None, default_port=8080, desc=''):
         parser = ArgumentParser(description=desc)
-        parser.add_argument('-p', '--port', type=int, default=default_port)
-        parser.add_argument('-b', '--bind', default='0.0.0.0')
-        parser.add_argument('-t', '--threads', type=int, default=4)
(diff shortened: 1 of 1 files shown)


marhop:
By the way, is it wise to use 0.0.0.0 as default for the `--bind` option? From a security point of view I would expect the wayback server to listen only on localhost by default (i.e., a default of 127.0.0.1 or localhost) and require explicit action (i.e., deliberately using something like `--bind 0.0.0.0`) to make it available on the network. I guess it depends on the use case though ...

ikreymer:
Thanks for the PR and sorry for delay in merging!

Re: `0.0.0.0` as default, that was actually added to address a specific usability issue. Since `8080` is sort of a common default port, a point of confusion was if there was already another app bound to 0.0.0.0. If pywb were to bind to 127.0.0.1, it would start up, but not actually be accessible in the browser, as the 0.0.0.0 binding took precedence. But perhaps it should still default to 127.0.0.1 for security reasons, as you mention. Instead, could add a temp check for binding to 0.0.0.0 to see if it is available, and warn otherwise.

Issue #343, opened 2018-06-13
Fuzzy matching bug with Flash files


despens:
This happens with pywb version 2.0.3 installed via pip on Python 3.6.5:

Pywb returns wrong data for certain URL requests, perhaps related to `.swf` (Flash) files.

To reproduce:

## Create WARC file

Directory structure:
```
files
├── path1
│   └── stuff
│       └── deep-down.html
└── path2
    ├── deep-down
    │   ├── deep-down.html
    │   ├── image.png
    │   └── what.swf
    └── index.html
```

The files are all have arbitrary contents.

Create a WARC file from this:

```bash
$ warcit -o http://www.website.co.br/~dinosaurs/t/ files -v
[DEBUG] Writing "http://www.website.co.br/~dinosaurs/t/path1/stuff/deep-down.html" (text/html) @ "2018-06-13T08:25:37Z" from "files/path1/stuff/deep-down.html"
[DEBUG] Writing "http://www.website.co.br/~dinosaurs/t/path2/index.html" (text/html) @ "2018-06-13T08:29:18Z" from "files/path2/index.html"
[DEBUG] Adding auto-index: http://www.website.co.br/~dinosaurs/t/path2/ -> http://www.website.co.br/~dinosaurs/t/path2/index.html
[DEBUG] Writing "http://www.website.co.br/~dinosaurs/t/path2/deep-down/image.png" (image/png) @ "2018-06-12T14:05:18Z" from "files/path2/deep-down/image.png"
[DEBUG] Writing "http://www.website.co.br/~dinosaurs/t/path2/deep-down/deep-down.html" (text/html) @ "2018-06-13T08:27:30Z" from "files/path2/deep-down/deep-down.html"
[DEBUG] Writing "http://www.website.co.br/~dinosaurs/t/path2/deep-down/what.swf" (application/x-shockwave-flash) @ "2018-06-13T08:39:31Z" from "files/path2/deep-down/what.swf"
[INFO] Wrote 6 resources to files.warc.gz
```

## Add to pywb

```bash
wb-manager add test files.warc.gz
wayback
```

## Access on localhost:

Requesting a path that partly matches that of any resource in the WARC will result in the flash file being served instead of a not found error.

![selection_999 725](https://user-images.githubusercontent.com/571494/41340801-7e31291c-6ef8-11e8-8c9b-e57ac82d76a0.png)

[...]

PR #362, opened 2018-08-06
Less aggressive fuzzy matching on mime type.

N0taN3rd:
## Description

This PR makes changes to the fuzzy matching in prefix mode by adding an additional match criteria to `mime` matches in `FuzzyMatcher.match_general_fuzzy_query`.
When a mime type match is being made also match on extension in order to be less aggressive when matching prefix matches.

## Motivation and Context

Fix for https://github.com/webrecorder/pywb/issues/343

## Types of changes

## Checklist:

### DIFF
--- a/pywb/warcserver/index/fuzzymatcher.py
+++ b/pywb/warcserver/index/fuzzymatcher.py
@@ -193,7 +193,8 @@ class FuzzyMatcher(object):
 
         # check mime
         mime = cdx.get('mime')
-        if mime and mime in self.default_filters['mimes']:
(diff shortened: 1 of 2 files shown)


ikreymer:
Expanded the additional criteria to include a prefix match and only varies by query. If matching a prefix query by extension or mime, ensure that the matched url is either:
- the requested url minus `?`
- the requested url + `?`

This is currently necessary as the default prefix skips the regex check, but must still exclude other paths, eg.

```
com,example)/file.png <- match this
com,example)/file/subpath/ <- skip this
com,example)/file.png?a=b <- match this
```

PR #365, opened 2018-08-17
Improves Docker experience

funkyfuture:
## Description

The provided Docker image is pinned to a fixed base image, its size and contents are reduced. An entrypoint script is added that allows to map the executing user and file ownership to user-defined ids.

## Motivation and Context

It started with not being able to record as `/webarchive` as it is owned by `root` while `wayback` runs as `archivist` who couldn't write there.

## Types of changes

## Checklist:

  - yet there's none for a deployment w/ Docker?


ikreymer:
Hi, apologies for delay in getting to this. Thanks for the PR, I think it makes sense but we'd like to make a couple changes as we improve the Dockerfile, such as adding the `ARG ` param back in so that base image is still configurable. Webrecorder users the base pywb image so we just need to do some more testing to make sure that it works well with alpine and python 3.6

ikreymer:
To avoid confusion, I've reset this PR as is and started new PR #446.

The entrypoint has been changed to be a bit different than what was proposed here,
for the following reasons:

- Ideally, we don't want to change the permissions on the volume, but rather match the user pywb is using to that of the volume. Changing the permissions can cause other issues for users and would like to avoid that. Instead, if the volume is mounted as root, we'll run pywb as root.
If mounted as another user, we create a new user to match the UID/GID. I've noticed that on different operating systems, the volume may be mounted as root or as the user running Docker, so this is designed to be as flexible as possible. Will hopefully also address webrecorder/webrecorder#650.

- The WR Dockerfile extends this image, so hence the additional requirements complexity.  To avoid maintaining/testing two different Dockerfiles, ideally that can continue to be the case.

- Also, set default python to 3.7, but with option to parametrize via ARG, if needed. This allows for configuring builds with different versions of python, but can default to 3.7 for now.

PR #366, opened 2018-08-17
Ensure WARC/0.18 metadata records with mime = text/anvl are not indexed for replay

N0taN3rd:
## Description

Added skipping of metadata records with mime = text/anvl to cdxindexer.py.

## Motivation and Context

Older web archiving software, especially those that generate WARCS using the WARC/0.18 specification, use the mime type text/anvl 
for there metadata records. 
This can make replay of such pages (mime text/html, et al) that also have corresponding metadata records with mime text/anvl impossible.

Fixes https://github.com/webrecorder/webrecorderplayer-electron/issues/63.

## Types of changes

## Checklist:

### DIFF
--- a/pywb/indexer/cdxindexer.py
+++ b/pywb/indexer/cdxindexer.py
@@ -38,6 +38,10 @@ import six
 
 #=================================================================
 class BaseCDXWriter(object):
+    # To ensure we do not index metadata mime types
+    # from older WARC specs (Heritrix 1.x) that collide with response records
+    METADATA_NO_INDEX_TYPES = ('text/anvl', )
+
     def __init__(self, out):
         self.out = codecs.getwriter('utf-8')(out)
         #self.out = out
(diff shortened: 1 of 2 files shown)


pywb/indexer/cdxindexer.py
```diff
@@ -50,11 +50,16 @@ def write(self, entry, filename):
...
-        if entry.record.rec_type == 'warcinfo':
+        if self._is_skipped(entry):
             return
 
         self.write_cdx_line(self.out, entry, filename)
 
+    def _is_skipped(self, entry):
+        if entry.record.rec_type == 'warcinfo':
+            return True
+        return entry.record.rec_type == 'metadata' and entry['mime'] == 'text/anvl'
```
ikreymer: Maybe should move this type to `METADATA_NO_INDEX_TYPES = ('text/anvl')` and do `entry['mime'] in METADATA_NO_INDEX_TYPES` in case we need to add other types later?

More broadly, maybe it makes sense to filter out a `metadata` record if it comes after a request/response pair and has the same url? The index is already checking for the pair, so maybe could also check if it encounters metadata if it was the same url? That should filter most of the Heritrix metadata records, I think..

N0taN3rd: Agree `METADATA_NO_INDEX_TYPES` class var is a good idea, updated PR to include this.

It might make sense to filter out `metadata` when it comes in that specific order tho Heritrix 3 creates a metadata record after info (containing outlinks).
That might be an issue tho if we run into this issue some more that approach will be the way to go.

Could also add an option to include metadata records when indexing where the default behavior is to exclude them.

pywb/indexer/cdxindexer.py
```diff
@@ -38,6 +38,10 @@ def json_encode(obj):
 
 #=================================================================
 class BaseCDXWriter(object):
+    # To ensure we do not index metadata mime types
+    # from older WARC specs (0.81) that collide with response records
```
ikreymer: WARC 0.18?  Can just say from Heritrix 1.x

PR #367, opened 2018-08-17
Ensure Pywb replays requests for JSON correctly when the original server who sent it said it was text/html   

N0taN3rd:
## Description

Add content sniffing to the html check of `_fill_text_type_and_charset` when the url ends with .json

## Motivation and Context

The page https://032c.com/fred-turner-silicon-valley-thinks-politics-doesnt-exist, makes a request for some JSON at URL

https://032c.com/fred-turner-silicon-valley-thinks-politics-doesnt-exist.json
but the server sends it over as text/html.

Fixes: https://github.com/webrecorder/webrecorder-bug-reports/issues/1457

## Types of changes

## Checklist:

### DIFF
--- a/pywb/rewrite/content_rewriter.py
+++ b/pywb/rewrite/content_rewriter.py
@@ -312,6 +312,7 @@ class StreamingRewriter(object):
 # ============================================================================
 class RewriteInfo(object):
     TAG_REGEX = re.compile(b'^\s*\<')
+    JSON_REGEX = re.compile(b'^\s*[{[][{"]')  # if it starts with this then highly likely not HTML
 
     JSONP_CONTAINS = ['callback=jQuery',
                       'callback=jsonp',
@@ -363,6 +364,10 @@ class RewriteInfo(object):
                 text_type = 'json'
 
         if (text_type and orig_text_type != text_type) or text_type == 'html':
+            if url.endswith('.json'):
+                buff = self.read_and_keep(56)
+                if self.JSON_REGEX.match(buff) is not None:
+                    return 'json', charset
             # check if default content_type that needs to be set
             new_mime = content_rewriter.default_content_types.get(text_type)
 
@@ -392,7 +397,7 @@ class RewriteInfo(object):
         # if html or no-content type, allow resolving on js, css,
         # or other templates
         if text_type == 'guess-text':
-            if not is_js_or_css and not mod in ('if_', 'mp_', ''):
+            if not is_js_or_css and mod not in ('if_', 'mp_', ''):
                 return None
 
         # if application/octet-stream binary, only resolve if in js/css content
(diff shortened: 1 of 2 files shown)


pywb/rewrite/content_rewriter.py
```diff
@@ -363,6 +364,10 @@ def _fill_text_type_and_charset(self, content_rewriter):
                 text_type = 'json'
 
         if (text_type and orig_text_type != text_type) or text_type == 'html':
+            if url.endswith('.json'):
+                buff = self.read_and_keep(56)
```
ikreymer: why 56?  :)

N0taN3rd: Using length 56 allows us to peak into the response body just enough to make the determination with confidence.

Since `b'^\s*\<'`was used previously with length 128 https://github.com/webrecorder/pywb/blob/2fbfe4424ade4e36e7127fde8643c4072e905afd/pywb/rewrite/content_rewriter.py#L411-L415

The regex  `^\s*[{[][{"]` would allows us to  detect two of the most common JSON patterns
- `{"data":{"complete"`  
- `["string", `

It could be expanded to cover `\{"data":\{"complete":true,` as well.

ikreymer: Yeah, but how'd you settle on 56? Should 64 or 128 be a better round buffer? :) 
The regex would just match the beginning, `["` or `{"`, which might be enough anyway to detect that its json and not html, so rest would just be for filtering whitespace.

We should probably have some tests for this also in `test_content_rewriter.py`

N0taN3rd: Agree we should have tests, wanted to make sure this PR was acceptable first given this is an edge case ;).

Length 56 get us: 
`b'{"data":{"complete":true,"storage":[{"key":"startup-data'`

Length 128 get us:
 `b'{"data":{"complete":true,"storage":[{"key":"startup-data","status":"complete","req":{"type":"startup-data","parser":"startup-dat'`

At length 56 we get the minimal amount of data needed to tell the difference between
`b'{"data":{"complete":true,"storage":[{"key":"startup-data'`
and 
`b'<script type="application/json">{"data":{"c`

N0taN3rd: Yes this edge case is specific to the site, only due to my not keeping a list of funkiness like this....
Right the JSON_REGEX only checks for starting with white space until `{` or `[` then stop.
Hence the chosen length of 56 could go shorter with the length as well / remove leading white space checking.

This is pattern is more common than not, usually with text/plain than HTML.
Also have HTML sent to script / response for images (img.src = url with response mime html).
Should we have a running issue with sites like this or no?

ikreymer: I think this is good for now, maybe can update the regex to be similar to jsonp_rewriter` to be consistent, and also account for possible starting comments? We can address other cases as needed.
Could start a separate issue for html for img.src, if that is a replay issue, though that may
not be as problematic as there should be no rewriting.

N0taN3rd: Comments are not allowed in JSON so that would be a dead giveaway it is not JSON :smiley:

ikreymer: Yep, you're right, that's part of the different between JSON and JSONP :)

PR #368, opened 2018-08-19
Add recorder option to filter source collection

eszense:
## Description
When using an aggregate collection or sequential fallback collection as the source, this change allow recording to be limited to pages fetched from certain child collection.

## Motivation and Context
The most probable use case is to limit recording to pages that are fetched from live but not from an existing collection. This can be accomplished by setting up an sequential fallback collection with an existing collection followed by live collection, then setting a recording filter. 

## Types of changes

## Checklist:

### DIFF
--- a/tests/test_record_replay.py
+++ b/tests/test_record_replay.py
@@ -173,19 +173,23 @@ class TestRecordFilter(HttpBinLiveTests, CollsDirMixin, BaseConfigTest):
                                      }
                      }
         super(TestRecordFilter, cls).setup_class('config_test_record.yaml', custom_config=rec_custom)
-
-    def test_init_and_rec(self):
         manager(['init', 'test-new'])
+        
+    def test_skip_existing(self):
         dir_name = os.path.join(self.root_dir, '_test_colls', 'test-new', 'archive')
         assert os.path.isdir(dir_name)
         res = self.testapp.get('/fallback/cdx?url=http://example.com/?example=1')
         assert res.text != ''
-        res = self.testapp.get('/fallback/cdx?url=http://httpbin.org/get?A=B')
-        assert res.text == ''
-
+        
         res = self.testapp.get('/test-new/record/mp_/http://example.com/?example=1')
         assert 'Example Domain' in res.text
         assert os.listdir(dir_name) == []
+    
+    def test_record_new(self):
+        dir_name = os.path.join(self.root_dir, '_test_colls', 'test-new', 'archive')
+        assert os.path.isdir(dir_name)
+        res = self.testapp.get('/fallback/cdx?url=http://httpbin.org/get?A=B')
+        assert res.text == ''
 
         res = self.testapp.get('/test-new/record/mp_/http://httpbin.org/get?A=B')
         assert res.json['args']['A'] == 'B'
--- a/tests/test_record_replay.py
+++ b/tests/test_record_replay.py
@@ -1,7 +1,7 @@
 from .base_config_test import BaseConfigTest, fmod, CollsDirMixin
 from pywb.manager.manager import main as manager
 from pywb.manager.autoindex import AutoIndexer
-from pywb.warcserver.test.testutils import to_path, HttpBinLiveTests
+from pywb.warcserver.test.testutils import to_path, HttpBinLiveTests, TEST_WARC_PATH, TEST_CDX_PATH
 
 import os
 import time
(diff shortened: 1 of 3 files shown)


ikreymer:
Thanks for the PR and the test and doc update!

We have thought about adding something like `/patch` endpoint to a collection that would record only the content that is missing.
It looks like your approach sets this up in a more manual way, but just uses the existing functionality!
The downside, is that you can only patch one collection at a time, since only one collection can be the recorder source currently.

If I understand correctly, the intended use case is to set up a new collection to patch only what's missing from an existing collection. An example might:

```
collections:
   patch-coll1:
       sequence:
            - name: coll1
               index_paths: ./collections/coll1/indexes
               archive_paths: ./collections/coll1/archives

             - name: live
                index_paths: $live

recorder:
    source_coll: patch-coll1
    source_filter: live
```

Then, you could create a new automatic collection, say `wb-manager init coll1-patch`
 and using then access `/coll1-patch/record/` would record only what's missing from `coll1`

To combine both, you'd then need to also have something like:

```
    combined:
       sequence:
            - name: coll1
               index_paths: ./collections/coll1/indexes
               archive_paths: ./collections/coll1/archives

             - name: coll1-patch
               index_paths: ./collections/coll1-patch/indexes
               archive_paths: ./collections/coll1-patch/archives
```

Then, you could replay the combined version with `/combined/`
Is this the use case that you're imagining?
Do you anticipate wanting to do this for multiple collection?
If so, the `/patch` endpoint may seem useful.
Do you want to have the patched content be separate from the original collection, as would happen here, or added to that collection?
The sequence and fallback pipeline is designed to be flexible, though the currently exposed options don't yet covert all the possible use cases.

tests/test_record_replay.py
```diff
@@ -154,3 +154,44 @@ def test_init_and_rec(self):
...
+    def test_init_and_rec(self):
+        manager(['init', 'test-new'])
+        dir_name = os.path.join(self.root_dir, '_test_colls', 'test-new', 'archive')
+        assert os.path.isdir(dir_name)
+        res = self.testapp.get('/fallback/cdx?url=http://example.com/?example=1')
+        assert res.text != ''
+        res = self.testapp.get('/fallback/cdx?url=http://httpbin.org/get?A=B')
+        assert res.text == ''
+
+        res = self.testapp.get('/test-new/record/mp_/http://example.com/?example=1')
```
ikreymer: Very minor change, could you split these two tests into separate functions, maybe `test_filtered_record_not_recorded` and `test_filtered_record_recorded` to indicate that first one is not recorded as its coming from example collection, and second one is recorded as its coming from live?

eszense:
Yes, my intended use case is actually to patch an existing one.

In fact, by configuring it like this:
```
collections:
   local_fallback:
       sequence:
            - name: local
               index_paths: ./collections/local/indexes
               archive_paths: ./collections/local/archives

             - name: live
                index_paths: $live

recorder:
    source_coll: local_fallback
    source_filter: live
```

A call to `/local/record/http://www.example.com` would add the page to the `local` collection if not already exist.

Having an `/patch` endpoint or `skip_duplicate: True` option in the recorder config would indeed be great. It's more intuitive and user-friendly. (I took me some time to figure out how to lay out a fallback sequence with filter to achieve the same). Also it would avoid the complexity of maintaining group of patches to the master collection (and replay with aggregate collections).

There is an implementation difficulty, however. By `/patch` , the end user would expect two consecutive `/patch` call to the same URL should only generate a single WARC record. However, the current proposed algorithm could not know it is a duplicate until the next reindexing. Using deduplication by Redis server (I see it is to be implemented) would be an alternative, but I think is quite undesirable having to maintain another Redis server. Or is there any other way? Maybe having an update-index-upon-warc-write? Would there be too much performance overhead?  

By `source_filter: live`,  I think it is more clear to the user that consecutive call to same URL can still generate multiple record. 

In short,

> Is this the use case that you're imagining? 
Yes. 

> Do you anticipate wanting to do this for multiple collection?
Not for my own usage, but I could imagine this would be useful for other users/developers

> If so, the /patch endpoint may seem useful.
Agree but it need some thought on how to properly implement `/patch`

> Do you want to have the patched content be separate from the original collection, as would happen here, or added to that collection?
Personally I would prefer to add it to the original collection, with the config setup above.

ikreymer:
Thanks for updating the test, sorry for delay. The PR looks good!

Yeah, I think the standard `/patch` endpoint functionality (like we have in Webrecorder) would be the preferred implementation and for your use, but this PR is a good option until we have a chance to implement that. 

> There is an implementation difficulty, however. By /patch , the end user would expect two consecutive /patch call to the same URL should only generate a single WARC record. However, the current proposed algorithm could not know it is a duplicate until the next reindexing. Using deduplication by Redis server (I see it is to be implemented) would be an alternative, but I think is quite undesirable having to maintain another Redis server. Or is there any other way? Maybe having an update-index-upon-warc-write? Would there be too much performance overhead?

Yep, you're absolute right that patching would require an in memory cache, eg using Redis or something else. This is what we use with the Webrecorder application to support patching, but perhaps there's a more standalone option that can be used in pywb itself. I think it could just be a simple in-memory cache that is flushed every time the re-indexing happens, so would only need to track urls recorded every 30 secs, for example.

We'll let you know if we have a chance to implement patching here!

PR #369, opened 2018-08-19
Favicon work

ikreymer:
## Description
Additional updates to favicon changes from #356. Favicon and title setting moved to default_banner.js from wb_frame.js to allow for easier overriding/customization.


ikreymer:
(Note: commit should not be squashed to maintain the original PR and updated changes as two separate commits).

Devhercule:
It is ok for me, it works

Can we see the commit in master (it appears in develop now)

ikreymer:
Yep, it'll be in the next, pywb 2.0.5 release, probably within next couple of weeks.

Issue #314, opened 2018-03-29
Running wayback behind a reverse proxy + https


bricas:
Is it possible to run wayback behind a reverse proxy that does https?

This is basically how our Kubernetes setup works, but my initial attempts didn't work right as it always wanted to push back a non-https url for assets and links.

Has this changed in 2.x release?

I believe I could probably deploy it using uwsgi w/ ssl suuport, but the app shouldn't need to be deployed with ssl directly.

ikreymer:
pywb uses the WSGI standard `wsgi.url_scheme` variable to determine the scheme, so if that's set to `https`, it should work

 If you're deploying with nginx and uwsgi and having nginx handle ssl, this can be done by adding the `UWSGI_SCHEME $scheme` param, eg, if pywb is running in uwsgi on port :8081, a config might be as follows:

```
    location / {
        include uwsgi_params;

        uwsgi_param UWSGI_SCHEME $scheme;
        uwsgi_pass localhost:8081;
    }
```

Hope this helps, I'm not sure if this is sufficient for Kubernetes, but would be curious to know. Perhaps should also be added to the pywb docs.

bricas:
I can confirm that if i switch from just `wayback` to `uwsgi uwsgi.ini` and add `nginx` on top using your sample config, i can get https urls returned.

This is not quite as trivial as I would've hoped (somehow getting `wayback` to return https urls via config or env variable) -- but it works.

ikreymer:
Ok, I've added a config setting you can use: `force_scheme: https` which will force the rewriting to use https with the standard `wayback` deployment. For production deployment, nginx + uwsgi is still the recommended.

bricas:
Thanks for the quick turn-around on this! Works great -- Thanks!

ikreymer:
@bricas Just wondering, would specifying the scheme via  [X-Forwarded-Proto](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto) or new [Forwarded](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded) headers, which are a bit more standard, work for your use case, or is the `force_scheme: https` still the best option?

bricas:
@ikreymer I believe our k8s setup sends the appropriate X-Forwarded-Proto header. So that would be the ideal, zero-config solution.

The force_scheme option is brute force, but works.

ikreymer:
[...]

PR #374, opened 2018-08-30
add sample Apache configuration

anarcat:
## Description

This patch updates the documentation to add a configuration snippet
for Apache.

## Motivation and Context

This configuration can be used when launching `wayback` in the default
configuration, which is useful to add stuff like access control,
authentication, or encryption without going through the trouble of
setting up a UWSGI proxy.

The `force_scheme` configuration wouldn't be necessary if pywb would
support the (de-facto) standard `X-Forwarded-Proto` header. This is
not, unfortunately, currently the case so the configuration cannot be
only on Apache's side. Note that this is similar to the concerns raised in #314

## Types of changes

## Checklist:

### DIFF
--- a/docs/manual/usage.rst
+++ b/docs/manual/usage.rst
@@ -189,3 +189,28 @@ See the `Nginx Docs <https://nginx.org/en/docs/>`_ for a lot more details on how
         }
     }
 
+Sample Apache Configuration
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The following Apache configuration snippet can be used to deploy pywb *without* uwsgi. A configuration with uwsgi is also probably possible but this covers the simplest case of launching the `wayback` binary directly.
+
+The configuration assumes pywb is running on port 8080 on localhost, but it could be on a different machine as well. The configuration can be updated to use HTTPS (as below), but the `force_scheme: https` option needs to be added to the `config.yml` file.
+
+.. code:: apache
+
+    <VirtualHost *:80>
+         ServerName proxy.example.com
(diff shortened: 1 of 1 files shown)


anarcat:
Also note that I've tried to use UWSGI with Apache, but ultimately failed. The procedure I follow, on Debian Buster, was to install the `uwsgi-core` and `uwsgi-plugin-python` packages, then run:

    uwsgi --plugin python uwsgi.ini

Then, on the server, I tried to configure the proxy with:

    ProxyPass "/" "uwsgi://localhost:8080/"
    ProxyPassReverse "/" "uwsgi://localhost:8080/"

It yielded a 500 internal server error, without anything particular in the logs, on the server or uwsgi side. Not sure what's up with that...

anarcat:
i'm not sure what to do with this - do you want me to make changes to the patch?

ikreymer:
@anarcat yes, I'm adding support for `X-Forwarded-Proto` header and deprecating `force_https` in #395 so can probably remove that and believe this should work:

```
RequestHeader set "X-Forwarded-Proto" expr=%{REQUEST_SCHEME}
``

anarcat:
okay, i've tweaked the docs to follow #395, so I guess this can be merged when the other lands.

anarcat:
and for the record, I just couldn't quite figure out how to make pywb work uwsgi with uwsgi. i don't have time right now to dig deeper into that, but it would be useful to have a better, step-by-step set of instructions on how to set that up instead of just delegating this to the (rather huge) uwsgi documentation. :)

thanks!

PR #375, opened 2018-09-05
rules: add fuzzy matching rule for vimeo, canonicalizing out a timest…

ikreymer:
## Description

Additional fuzzy matching rule for vimeo videos, canonicalizing the variable hmac/timestamp portion of url.


N0taN3rd:
LGTM Merging

PR #376, opened 2018-09-05
Encoding Fix

ikreymer:
## Description
Additional fixes from bug introduced in #361, which could cause urls to be reencoded with the wrong encoding. The assumption was that all unicode urls will be %-encoded. However, if a url is not rewritten, such as an anchor, it would not be %-encoded.

Now checking if rewritten url is the same, and then returning the original value.

Also, optimizing for utf-8: if content known to be utf-8, use that encoding for decoding the stream for rewriting. Otherwise, default to decoding as latin-1 (ignore encoding), and re-encode only the urls as utf-8.
 
## Types of changes

## Checklist:

### DIFF
--- a/pywb/rewrite/content_rewriter.py
+++ b/pywb/rewrite/content_rewriter.py
@@ -277,7 +277,7 @@ class StreamingRewriter(object):
         self.first_buff = first_buff
 
     def __call__(self, rwinfo):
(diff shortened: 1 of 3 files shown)


N0taN3rd:
LGTM!

PR #377, opened 2018-09-06
encoding fix: additional fix to #376 for banner encoding:

ikreymer:
- if no encoding is detected, don't default to utf-8
- if no encoding known, encode banner as 'ascii' with 'xmlcharrefreplace', converting to xml entities
- tests: add tests for rewriting with no known encoding

### DIFF
--- a/pywb/rewrite/content_rewriter.py
+++ b/pywb/rewrite/content_rewriter.py
@@ -153,9 +153,9 @@ class BaseContentRewriter(object):
                 except:
                     pass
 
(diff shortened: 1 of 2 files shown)


N0taN3rd:
Encoding spot checks look good, merging
http://pelicanbomb.com/events/archives/2011   :heavy_check_mark: 
https://en.wikipedia.org/wiki/Atlantic_Ocean  :heavy_check_mark: 
http://esukhia.org/                           :heavy_check_mark:

PR #378, opened 2018-09-06
Ensure that Preservation workers deferredSheetExtraction is guarded against null CSSRuleList

N0taN3rd:
## Description

Added check for rules to be null to the noop case in  deferredSheetExtraction (edge case)
Ensures that the CSS rule extraction from style tags checks both probably properties that the CSSRuleList can be on (cssRules or rules)

See https://www.youtube.com/ in FF.

## Types of changes

## Checklist:


N0taN3rd:
@ikreymer yes it would and done

Issue #371, opened 2018-08-20
Support operating in proxy mode with preservation worker  


N0taN3rd:
Because we are not intercepting element addition in proxy mode, we will need to check periodically to see if `img.srcset` or new media rules have been introduced into the document.

This will require a more involved de-duplication strategy to avoid not using mutation observers

N0taN3rd:
closed via PR https://github.com/webrecorder/pywb/pull/379 merger

PR #379, opened 2018-09-07
Preservation Worker Proxy Mode 

N0taN3rd:
## Description

Implemented preservation worker for proxy mode:
- Split wombat and preservation worker into two files (proxy mode and non-proxy mode)
- Started root config file control over including wombat and preservation worker in proxy or non-proxy mode
- Added additional proxy mode + preservation worker only route for fetching the preservation worker code nicely for CORS
- head_insert.html: only include wombat in proxy mode when env.pywb_proxy_magic exists and wombat_mode states we should include it
- templateview.py: when in proxy mode HeadInsertView.create_insert_function calculates the wombat_mode proxy mode is operating in, `w = use wombat`, `p = use preservation worker` and an empty string means do not use either.
- wombatProxyMode.js: preservation worker is used when wbinfo.wombat_mode includes p and the subset of the full wombat when it includes `w`
- cli.py: Added additional flags to BaseCLI for controlling wombat `--proxy-with-wombat` and preservation worker `--proxy-with-preserve-worker` inclusion

Updated tests:
- test_wbrequestresponse.py: added tests covering constructor defaults, _init_derived, options_response, json_response, encode_stream, text_stream
- test_auto_colls.py: fixed broken test test_more_custom_templates, reason using ujson now not json so spacing was off
- test_proxy.py: updated existing tests to reflect splitting wombat into proxy and non-proxy mode, added tests covering preservation worker specific endpoints in proxy mode
- test_cli.py: updated to properly test the cli with these changes

Fully documented via docstrings:
- cli.py
- frontendapp.py
- templateview.py
- wbrequestresponse.py

Removed duplicate addons key in .travis.yml
Added ultrajon dep to tests_require in setup.py to reflect its usage by wbrequestresponse.py

## Motivation and Context

Fixes https://github.com/webrecorder/pywb/issues/371
## Types of changes

## Checklist:

### DIFF
--- a/pywb/apps/frontendapp.py
+++ b/pywb/apps/frontendapp.py
@@ -6,7 +6,7 @@ from werkzeug.exceptions import HTTPException, NotFound
 from werkzeug.wsgi import pop_path_info
 from six.moves.urllib.parse import urljoin
 from six import iteritems
-
+from warcio.statusandheaders import StatusAndHeaders
 from warcio.utils import to_native_str
 from wsgiprox.wsgiprox import WSGIProxMiddleware
 
@@ -33,6 +33,16 @@ import logging
 
 # ============================================================================
 class FrontEndApp(object):
+    """Orchestrates pywb's core Wayback Machine functionality and is comprised of 2 core sub-apps and 3 optional apps.
+
+    Sub-apps:
+      - WarcServer: Serves the archive content (WARC/ARC and index) as well as from the live web in record/proxy mode
+      - RewriterApp: Rewrites the content served by pywb (if it is to be rewritten)
+      - WSGIProxMiddleware (Optional): If proxy mode is enabled, performs pywb's HTTP(s) proxy functionality
+      - AutoIndexer (Optional): If auto-indexing is enabled for the collections it is started here
+      - RecorderApp (Optional): Recording functionality, available when recording mode is enabled
+    """
+
     REPLAY_API = 'http://localhost:%s/{coll}/resource/postreq'
     CDX_API = 'http://localhost:%s/{coll}/index'
     RECORD_SERVER = 'http://localhost:%s'
@@ -45,6 +55,10 @@ class FrontEndApp(object):
     PROXY_CA_PATH = os.path.join('proxy-certs', 'pywb-ca.pem')
 
     def __init__(self, config_file='./config.yaml', custom_config=None):
+        """
+        :param str config_file: Path to the config file
+        :param dict custom_config: Dictionary containing additional configuration information
+        """
         self.handler = self.handle_request
         self.warcserver = WarcServer(config_file=config_file,
                                      custom_config=custom_config)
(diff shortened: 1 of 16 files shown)


pywb/templates/head_insert.html
```diff
@@ -24,10 +24,17 @@
   wbinfo.coll = "{{ coll }}";
   wbinfo.proxy_magic = "{{ env.pywb_proxy_magic }}";
   wbinfo.static_prefix = "{{ static_prefix }}/";
+{% if env.pywb_proxy_magic %}
+  wbinfo.wombat_mode = "{{ wombat_mode }}";
```
ikreymer: The template already has access to the proxy config, so I think this can probably just be simplified as:

```
wbinfo.useWombat = {{ "true" if config.proxy.use_wombat else "false" }};
wbinfo.usePreserveWorker  =  {{ "true" if config.proxy.use_preserve_worker else "false" }};
```

pywb/static/wombatProxyMode.js
```diff
@@ -0,0 +1,357 @@
...
+                }
+            }, false);
+        }
+    }
+
+    if (wbUseAAWorker) {
+        initAutoArchivingWorker();
+    }
+
+    if (wbUseWombat) {
```
ikreymer: Hm, do we want to support a mode where there is 'Auto-Archiving' worker but not regular wombat?
Maybe for simplicity, there should just be 3 options:
- No rewriting
- Minimal proxy wombat
- Minimal proxy wombat + auto archiving worker
?

ikreymer:
Also, should probably standardize on the name for this system.. we had preserve worker, archive worker.. Since its primarily useful for recording extra content while running in the background and useful with recording mode specifically, perhaps
'BgRecorder', 'BgProxyRecorder' and 'BgRecordingWorker'.. There's a few places where its not quite consistent yet, but pretty close. Just thinking about how this system should be called overall :)

pywb/static/wombatProxyMode.js
```diff
@@ -0,0 +1,359 @@
...
+        function AutoFetchWorker() {
+            if (!(this instanceof AutoFetchWorker)) {
+                return new AutoFetchWorker();
+            }
+            this.checkIntervalCB = this.checkIntervalCB.bind(this);
+            if (isTop) {
+                // Cannot directly load our worker from the proxy origin into the current origin
+                // however we fetch it from proxy origin and can blob it into the current origin :)
+                var self = this;
+                fetch(wb_info.static_prefix + 'autoFetchWorkerProxyMode.js')
```
ikreymer: Small request here: could the prefix be made overridable, eg:
```
var wbAutoFetchWorkerPrefix = (wb_info.auto_fetch_worker_prefix || wb_info.static_prefix) + 'autoFetchWorkerProxyMode.js'
```
In WR, will route it to a different prefix to add cors headers..

pywb/apps/frontendapp.py
```diff
@@ -431,14 +575,110 @@ def proxy_route_request(self, url, environ):
         """
         return self.proxy_prefix + url
 
+    def proxy_auto_fetch_worker(self, env, *args):
```
ikreymer: This function is no longer needed now with the static prefix routing, right?

N0taN3rd: done

pywb/static/wombatProxyMode.js
N0taN3rd: done

PR #381, opened 2018-09-18
Add setTimeout, setInterval, and window.origin overrides to wombat.js

N0taN3rd:
## Description

wombat.js:
- Added override for setTimeout and setInterval because [setTimeout|set…Interval]('document.location.href = "xyz.com"', time) is legal and used
- Added override for window.origin (https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/origin) available in Chrome 59+ and FF 54+

## Motivation and Context

URL for `setTimeout` and `setInterval`: https://guides.library.upenn.edu/c.php?g=870924
- Site tracks link clicks and defers browser navigation by doing `setTimeout('document.location.href = "xyz.com"', time)` once they have done their tracking work 

## Types of changes

## Checklist:

### DIFF
--- a/pywb/static/wombat.js
+++ b/pywb/static/wombat.js
@@ -350,7 +350,7 @@ var _WBWombat = function($wbwindow, wbinfo) {
                 return url;
             }
 
-            // relative collection 
+            // relative collection
             if ((url.indexOf(wb_rel_prefix) == 0) && (url.indexOf("http") > 1)) {
                 var scheme_sep = url.indexOf(":/");
                 if (scheme_sep > 0 && url[scheme_sep + 2] != '/') {
@@ -1687,6 +1687,19 @@ var _WBWombat = function($wbwindow, wbinfo) {
     }
 
     //============================================
+    function wrapScriptTextJsProxy(scriptText) {
+        return 'var _____WB$wombat$assign$function_____ = function(name) {return (self._wb_wombat && ' + 'self._wb_wombat.local_init &&self._wb_wombat.local_init(name)) || self[name]; };\n' +
+            'if (!self.__WB_pmw) { self.__WB_pmw = function(obj) { return obj; } }\n{\n' +
+            'let window = _____WB$wombat$assign$function_____("window");\n' +
+            'let self = _____WB$wombat$assign$function_____("self");\n' +
+            'let document = _____WB$wombat$assign$function_____("document");\n' +
+            'let location = _____WB$wombat$assign$function_____("location");\n' +
+            'let top = _____WB$wombat$assign$function_____("top");\n' +
+            'let parent = _____WB$wombat$assign$function_____("parent");\n' +
+            'let frames = _____WB$wombat$assign$function_____("frames");\n' +
+            'let opener = _____WB$wombat$assign$function_____("opener");\n' + scriptText + '\n\n}';
+    }
+
     function rewrite_script(elem) {
         if (elem.getAttribute("src") || !elem.textContent || !$wbwindow.Proxy) {
             return rewrite_attr(elem, "src");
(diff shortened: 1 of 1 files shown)


pywb/static/wombat.js
```diff
@@ -2856,7 +2856,72 @@ var _WBWombat = function($wbwindow, wbinfo) {
         wb_funToString.apply = orig_apply;
     }
 
-
+    function initTimeoutIntervalOverrides($wbwindow) {
+        // because [setTimeout|setInterval]('document.location.href = "xyz.com"', time) is legal and used
+        // declare the beginning wrapper string first so we do not redeclare it multiple time in our override (closure)
+        var wrapper = 'var _____WB$wombat$assign$function_____ = function(name) {return (self._wb_wombat && self._wb_wombat.local_init &&self._wb_wombat.local_init(name)) || self[name]; };\n' +
```
ikreymer: To avoid duplication, could this be combined with what's happening in `rewrite_script()`?
Maybe we can factor out the wrapper creation into a separate function, `wrap_script_js_proxy()` that computes prefix, like in rewrite_script on first use,
and then applies it to a string? Then, this function and rewrite_script() could use wrap_script_js_proxy()?

A little more refactoring, but I think will be worth it in keep this more maintainable :)

pywb/static/wombat.js
```diff
@@ -2856,7 +2856,72 @@ var _WBWombat = function($wbwindow, wbinfo) {
...
+        if ($wbwindow.setTimeout && !$wbwindow.setTimeout.__$wbpatched$__) {
+            var originalSTO = $wbwindow.setTimeout;
+            $wbwindow.setTimeout = function () {
+                // strings are primitives with a prototype or __proto__ of String depending on the browser
+                var rw = arguments[0] != null && Object.getPrototypeOf(arguments[0]) === String.prototype;
+                // do not mess with the arguments object unless you want instant de-optimization
+                var args = rw ? new Array(arguments.length) : arguments;
+                if (rw) {
+                    if ($wbwindow.Proxy) {
+                        args[0] = wrapper + arguments[0] + '\n}';
```
ikreymer: See above, ideally this could just be `wrap_script_js_proxy(arguments[0])`

pywb/static/wombat.js
```diff
@@ -2856,7 +2856,72 @@ var _WBWombat = function($wbwindow, wbinfo) {
...
+                    } else {
+                        args[0] = arguments[0].replace(/\blocation\b/g, "WB_wombat_$&");
+                    }
+                    for (var i = 1; i < arguments.length; ++i) {
+                        args[i] = proxy_to_obj(arguments[i]);
+                    }
+                }
+                // setTimeout does not require its this arg to be window so just in case someone got funky with it
+                if (originalSTO.__WB_orig_apply) {
+                    return originalSTO.__WB_orig_apply(null, args)
```
ikreymer: Think we just need to add proxy_to_obj here as well and then it'll work. Seems safer than null:
`originalSTO.__WB_orig_apply(proxy_to_obj(this), args)`

pywb/static/wombat.js
```diff
@@ -3663,13 +3735,12 @@ var _WBWombat = function($wbwindow, wbinfo) {
...
             override_func_first_arg_proxy_to_obj($wbwindow.Document, "createTreeWalker");
 
-            override_func_this_proxy_to_obj($wbwindow, "setTimeout");
-            override_func_this_proxy_to_obj($wbwindow, "setInterval");
             override_func_this_proxy_to_obj($wbwindow, "getComputedStyle", $wbwindow);
             //override_func_this_proxy_to_obj($wbwindow.EventTarget, "addEventListener");
             //override_func_this_proxy_to_obj($wbwindow.EventTarget, "removeEventListener");
 
             override_apply_func($wbwindow);
+            initTimeoutIntervalOverrides($wbwindow);
```
ikreymer: To avoid duplication, lets just run the same code for setTimeout and setInterval:
```
initTimeoutIntervalOverrides($wbwindow, "setTimeout");
initTimeoutIntervalOverrides($wbwindow, "setInterval");
```

pywb/static/wombat.js
```diff
@@ -3402,12 +3469,17 @@ var _WBWombat = function($wbwindow, wbinfo) {
...
-        } else if (prop == "_WB_wombat_obj_proxy") {
-            return obj._WB_wombat_obj_proxy;
+        switch (prop) {
+            case '__WBProxyRealObj__':
+                return obj;
+            case 'location':
+                return obj.WB_wombat_location;
+            case '_WB_wombat_obj_proxy':
+                return obj._WB_wombat_obj_proxy;
+            case 'setTimeout':
```
ikreymer: Don't think this override is needed, as it defaults to `obj[prop]` anyway, right?
It should just return the bound function, which should work as expected.

N0taN3rd: Yes the default return value is `obj[prop]` (bound to correct this if function) but since we added the setTimeout and setInterval overrides which handles binding the this arg. 
This statement acts as an early out

ikreymer: Yeah, but do we need the early out though? We're not doing it for other overridden functions on window like postMessage, addEventListener, etc... should we just remove it to be consistent? Just thinking if coming back to this later will think that it is needed for some reason for setTimeout/setInterval specifically.. All the other cases are required.

pywb/static/wombat.js
ikreymer: Ah wow, nice find/docs of the issue! I wonder if something else is the cause of this exact issue, but should be good for now. Something to keep in mind if we see this again.

Issue #408, opened 2018-11-13
do not copy files into archive


anarcat:
## Is your feature request related to a problem? Please describe.

I find it difficult to use pywb on large datasets because the files are *copied* into the collections instead of just "referenced" there.

## Describe the solution you'd like

When I add a file to a collection, it should be just treated as if it's in the collection somehow, without having to copy gigantic files around.

## Describe alternatives you've considered

I have looked for options in the `wb-manager` program to see if something could fit the bill, particularly a way to symlink or hardlink files around. Haven't found anything. I also looked at the [auto-indexer](https://pywb.readthedocs.io/en/latest/manual/usage.html#dynamic-collections-and-automatic-indexing) but i'm not sure how that works.

traverseda:
On copy-on-wire filesystems like btrfs it would be nice if we could use btrfs's reflinks.

PR #409, opened 2018-11-13
WIP: allow symlinkinkg and hardlinking files instead of just copying

anarcat:
## Description

This allows users to manage collections of large WARC files without duplicating space. Hardlinks are used instead of symlinks to reflect the original mechanism, where the file is copied (so it can be safely removed from the source). If we used symlinks, we would break that expectation which could lead to data loss.

Inversely, hardlinks can lead to data loss as well. For example, pywb could somehow edit the file, which would modify the original as well. But we assume here pywb does not modify the file, and each side of the hardlink can have their own permissions to ensure this (or not) as well.
    
Closes: #408

## Types of changes

## Checklist:

### DIFF
--- a/pywb/manager/manager.py
+++ b/pywb/manager/manager.py
@@ -126,8 +126,10 @@ directory structure expected by pywb
             elif method == 'symlink':
                 os.symlink(filename, os.path.join(self.archive_dir,
                                                   os.path.basename(filename)))
-            else:
+            elif method == 'copy':
                 shutil.copy2(filename, self.archive_dir)
+            else:
+                raise NotImplementedError('unknown method name: %s' % method)
             full_paths.append(os.path.join(self.archive_dir, filename))
 
         self._index_merge_warcs(full_paths, self.DEF_INDEX_FILE)
--- a/pywb/manager/manager.py
+++ b/pywb/manager/manager.py
@@ -108,7 +108,7 @@ directory structure expected by pywb
                    'To create a new collection, run\n\n{1} init {0}')
             raise IOError(msg.format(self.coll_name, sys.argv[0]))
 
-    def add_warcs(self, warcs):
+    def add_warcs(self, warcs, hardlink=False):
         if not os.path.isdir(self.archive_dir):
             raise IOError('Directory {0} does not exist'.
                           format(self.archive_dir))
@@ -116,9 +116,16 @@ directory structure expected by pywb
         full_paths = []
         for filename in warcs:
             filename = os.path.abspath(filename)
-            shutil.copy2(filename, self.archive_dir)
+            if hardlink:
+                os.link(filename, os.path.join(self.archive_dir,
+                                               os.path.basename(filename)))
+            else:
+                shutil.copy2(filename, self.archive_dir)
             full_paths.append(os.path.join(self.archive_dir, filename))
-            logging.info('Copied ' + filename + ' to ' + self.archive_dir)
+            logging.info('%s %s to %s',
+                         hardlink and 'Linked' or 'Copied',
+                         filename,
+                         self.archive_dir)
 
         self._index_merge_warcs(full_paths, self.DEF_INDEX_FILE)
 
@@ -357,12 +364,14 @@ Create manage file based web archive collections
     # Add Warcs
     def do_add(r):
         m = CollectionsManager(r.coll_name)
-        m.add_warcs(r.files)
+        m.add_warcs(r.files, r.hardlink)
 
     addwarc_help = 'Copy ARCS/WARCS to collection directory and reindex'
     addwarc = subparsers.add_parser('add', help=addwarc_help)
     addwarc.add_argument('coll_name')
     addwarc.add_argument('files', nargs='+')
+    addwarc.add_argument('--hardlink', '-l', action='store_true',
+                         help='hardlink files into storage instead of copying')
     addwarc.set_defaults(func=do_add)
 
     # Reindex All
--- a/pywb/manager/manager.py
+++ b/pywb/manager/manager.py
@@ -117,7 +117,7 @@ directory structure expected by pywb
         for filename in warcs:
             filename = os.path.abspath(filename)
             logging.info('%s %s to %s',
-                         method,
+                         method.title(),
                          filename,
                          self.archive_dir)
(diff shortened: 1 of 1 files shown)


despens:
Hey, if your WARC files are already present somewhere else in a structured way, you might have success with configuring multiple archive paths, see https://pywb.readthedocs.io/en/latest/manual/configuring.html#archive-paths

anarcat:
The problem with that approach is that this expects a certain layout in the filesystem. Right now WARC files are stored like this:

    archive/example.com-YYYY-MM-DD-HASH/fooo-NNNN.warc.gz

Where:

 * `YYYY-MM-DD` is the date of the crawl
 * `HASH` is a unique hash
 * `NNNN` is an incrementing number (e.g. `0000` for the first WARC file, `0001` for the second one, files are split on 5GB boundaries)

This layout is commonly used by grab-site and archivebot and other crawlers. It does not match the expectations of pywb, including custom archive paths, which still look in `collection/<coll name>/example.warc.gz`.

So I don't think that's a sufficient solution to my problem.

JustAnotherArchivist:
I can't speak for the technical implementation (although it looks good to me), but I'd definitely appreciate anything that allows using existing WARCs – whatever file structure they may be in – in pywb without having to make any copies. Collections of WARCs can be huge, and storing them twice is wasteful and in some cases not even possible due to space constraints.

Of course, it would be possible to create a separate directory with the expected structure, make hardlinks or symlinks in there, and then use that as another archive path. But from a usability perspective, it would be much better in my opinion if pywb/wb-manager simply had an option that can take care of that directly (like the one proposed in this PR) instead of having to do that manually or with helper scripts.

ikreymer:
Thanks for suggesting this, I agree that this should be supported, but not sure that symlinking/hardlinking is the way to go. As @N0taN3rd, this would complicate Windows support and would potentially make the setup more brittle.

pywb is close to supporting what you want with external paths, but unfortunately its not automated.

It seems like the best option would be to support a per-collection overrides, say `overrides.yaml` which can set a list of one or more paths to directories that contain WARC files, ex:

collections/external-data/overrides.yaml:

```
archive_paths:
  - /path/to/warcs/
  - /path/to/more/warcs/
  - /path/to/specific/warcfile.warc.gz
```

Then, instead of the local `archives` directory, pywb will look for works in those directories instead.
This can also work with auto-indexing, so any time a new WARC is added to those directories, it can be indexed automatically. pywb can already check and index subdirectories, so you can use whatever structure you'd like in the external directory.

Of course, the overrides.yaml could also be managed with something like:
```
wb-manager add-external-path <coll> <path>
wb-manager remove-external-path <coll> <path>
```

If you wanted to add only a specific WARC file in a directory instead of all WARC, that too can be supported by specifying the file path instead of a directory (eg. `/path/to/some-warcs/warcfile.warc.gz`)

JustAnotherArchivist:
Where would the CDX files be stored with that setup?

Regarding Windows support, is that a hard requirement or could that feature simply be unavailable if not supported by the OS, the Python version, or the configuration (only admins can create symlinks on Windows according to the Python docs)?

ikreymer:
> Where would the CDX files be stored with that setup?

By default, still in the `indexes` directory, although can provide an override for that also, if needed.

> Regarding Windows support, is that a hard requirement or could that feature simply be unavailable if not supported by the OS, the Python version, or the configuration (only admins can create symlinks on Windows according to the Python docs)?

Ideally, the external directory feature would be available on all platforms.

anarcat:
what I hear from the various comments so far is this:

 * LGTM, some fix like this is necessary
 * this can be done by hand
 * add support for moving files (`os.move`, presumably, although that fails if we cross FS boundaries)
 * external paths are close to answering that requirement, but would require a new config file, maybe modifiable with a new set of commands
 * concerns about Windows support
 * pointers to which docs need updating (`usage.rst`, thanks!)

I understand where you are coming from: multi-platform compatibility is important, and there are existing features which might fit this requirement.

however, i would argue that "a bird in the mouth is better than two in the bush": I have a working patch to workaround a real scalability issue with pywb, right now. it might not work that effectively on Windows, but I want to point out that both `os.link` and `os.symlink` are actually supported on Windows, at least since Vista. so i don't think it's as much a blocker as people would tend to believe

the proposals to automate editing of the YAML file seem to be an entirely different approach, one that would require much more changes to the documentation and seem to me like feature creep. i just want to copy files lightly in the archive, not redesign how the entire YAML configuration system works! :) if this is the approach you want to take, I'm not sure I can help since I would need to dive again deeper in the internals of pywb, which might mean this would never be done at all. ;)

so to move this forward, I would propose that we keep on following the approach I proposed here. this would mean adding tests for the functionality and documentation. i would be happy to push that forward, if the proposal is accepted, otherwise I'm afraid I won't be able to provide a solution to #408 myself going forward.

have a nice day!

PS: the travis test failure here does not seem related to the patch, you might want to look into that... i'll re-trigger the build to see if it works better now.

ikreymer:
hi @anarcat apologies for the delay -- was away on vacation and neglected to respond earlier!

I reread your comment and thought about it more, and have considered our current work. Since we don't have the bandwidth to implement the config-based approach now, and it could still be done at a later time, I think you're right and we should add this solution, even if it is not cross-platform as it will help your (and possibly others') use cases. This solution is a simple change to the wb-manager while the config option would be a much more extensive change, as you've mentioned.

To proceed, could you add:
- tests (and mark them with something like `@pytest.mark.skipif(sys.platform == 'win32',
                    reason="does not run on windows")` to skip windows)
Existing tests for manager are mostly in `./tests/test_auto_colls.py` and `./tests/test_redirects.py` is another example. A test module similar to those (but simpler probably!) should be good.

- docs in `usage.rst`, and include some tradeoffs between using the hardlink vs symlink vs copy approaches.

And we'll try to merge it in for next release! Thanks again!

(Yes, the travis-ci issue is/was unrelated, we're looking at that)

anarcat:
awesome! i'm not sure I'll have time to do this before the next year (and I'm don't mind at all if someone else beats me to it), but hopefully I'll be able to come back to this soon-ish.

traverseda:
Any progress with this? It's something I'm finding problematic at the moment.

Issue #296, opened 2018-02-08
New calendar view doesn't work for URL searches


despens:
The new calendar view only works for date search, not URL search or date+URL search. When using URL search, it shows very confusing results.

despens:
The calendar should only be shown on searches that do not involve a wildcard in the URL part.

N0taN3rd:
This functionality was added in PR #421

PR #421, opened 2018-11-21
Improved Query Interface and Result viewing

N0taN3rd:
## Description

Added more appealing styling to the error, index, not-found, query, and  search templates (see the screen shots section).
Updated the included jquery and boostrap static files to jQuery v3.3.1 and Bootstrap v4.1.3 to facilitate the changes.

Exposed cdx api's through the query html page
- from, to
- matchType
- filter

Reworked query.js to know the difference between date search and advanced searching.

## Motivation and Context

Ref https://github.com/webrecorder/pywb/issues/296 and https://github.com/ukwa/ukwa-pywb/issues/32

### Home Page
pywb-home-page

### Default Search
default-search

### Default Search Results
url-search-results

### Advanced Search
searchpageadvanced

### Advanced Search Results
advancedsearchresults

## Types of changes

## Checklist:


ikreymer:
This is really neat! It looks like it doesn't support the conventional calendar query, though:
`/*/<url>` and also would it be possible to support `/*/<url>*` (assuming the url does not have a '*')
and possible `/from-to/<url>`? This is all just for backwards compatibility with existing conventions.

Also, if the query could match the cdx server api as much as possible, just to make it consistent?
Eg. maybe just have url, matchType, filter, etc.. be top-level query arguments that are just passed to cdx server api, and put anything extra into `humanFilter`, to indicate those are extra params for the UI only?

Overall, this is a really big improvement!

PR #426, opened 2018-11-30
Force absolute URLs when rewriting link href's to ensure font files are decoded correctly when response mime type does not indicate font

### LINKED ISSUE #424: Rewriting: leaving URLs relative in link[rel='stylesheet'] (/something/style.css) causes font files not be parsed when response mime type  does not indicate font
## Expected behavior

Fonts should be loaded

## What actually happened

LW Page: http://www.csg.gob.mx/
contains in the head
```html
<link rel="stylesheet" href="css/css/estilos.css">
```

pywb live this is get resolved by the browser as 
```html
<link rel="stylesheet" href="http://localhost:8080/live/mp_/http:/www.csg.gob.mx/css/css/estilos.css">
```
 
Because the css files import fonts and the font URLs are also relative they get resolved with the `mp_` modifier thus causing the font decode failure 

Font URL: <pywb prefix>/mp_/http://www.csg.gob.mx/fonts/glyphicons-halflings-regular.woff2
Response headers:
```http
Content-Location: http://localhost:8080/live/20181130232138mp_/http://www.csg.gob.mx/fonts/glyphicons-halflings-regular.woff2
Content-Security-Policy: default-src 'unsafe-eval' 'unsafe-inline' 'self' data: blob: mediastream: ws: wss: ; form-action 'self'
Content-Type: text/html;charset=UTF-8
```
contents:
```
wOF2Fl±\F	M?FFTM `r
©$åe6$t0 "Q?webfe5ìó@ Â?þ¾
Úö ¬töÿ ÁÇØÜ,3+2q
ËF®YOì&>±é±bÒm¤5ZæH$±ªY½{ìH	jdÕü²%÷¹Ù§y"§÷ö×+@¹]½«eû{úÒûv Nc)»nùÉ?~?è¤hòÿ_ç&iÆÊÈÑÁ¤?º>üå^K ûv´-cÛ1íô2K áy·õ,'nä«à(ð3EwiàB &©ÿéÎT´lh1<!-- WB Insert -->
<script>
  wbinfo = {};
  wbinfo.top_url = "http://localhost:8080/live/http://www.csg.gob.mx/fonts/glyphicons-halflings-regular.woff2";
  // Fast Top-Frame Redirect
  if (window == window.top && wbinfo.top_url) {
    var loc = window.location.href.replace(window.location.hash, "");
    loc = decodeURI(loc);
```
 

## Browser

All of them

N0taN3rd:
## Description

Ensure that we rewrite to abs because if we do not and encounter a super relative URL (href='css/stylesheet.css') under a page rooted
under the mp_ modifier and the response from the server does not indicate the right mime-type, fonts et al can get messed up

## Motivation and Context

fixes #424 

## Types of changes

## Checklist:

### DIFF
--- a/pywb/rewrite/html_rewriter.py
+++ b/pywb/rewrite/html_rewriter.py
@@ -455,8 +455,11 @@ class HTMLRewriterMixin(StreamingRewriter):
 
         elif rel == 'stylesheet':
             rw_mod = 'cs_'
-
-        return self._rewrite_url(attr_value, rw_mod)
+        # ensure that we rewrite to abs because if we do not and encounter a
+        # super relative URL (href='css/stylesheet.css') under a page rooted
+        # under the mp_ modifier and the response from the server does not
+        # indicate the right mime-type, fonts et al can get messed up
+        return self._rewrite_url(attr_value, rw_mod, True)
 
     def _set_parse_context(self, tag, tag_attrs):
         # special case: script or style parse context


ikreymer:
After discussion, going with #428 as an alternative, more general solution and to avoid rewriting relative links as absolute, which can cause more issues

Issue #434, opened 2019-01-23
Encoding issues with many popular sites


pirate:
## Describe the bug
When using `wayback` in both `--proxy` and `--proxy-record` mode, many sites seem to display gibberish, possibly because of an incorrect encoding being forced somewhere.

Working sites:
 - https://example.com
 - https://reddit.com
 - https://github.com
 - and many more...

Broken sites:
 - https://google.com
 - https://facebook.com
 - https://www.cloudflare.com
 - and many more...

I'm not sure what makes some sites work and not others.

## Steps to reproduce the bug
1. Use Python 3.7.2 and either pywb v2.1.0 or pywb-2.2.0.dev0 (latest `develop` branch)
2. Using an empty `config.yaml`, run:
```bash
wb-manager init demo
wayback --proxy-record --proxy demo
```
3. Open a page in a browser:
```bash
google-chrome --proxy-server=http://localhost:8080 --ignore-certificate-errors --disable-web-security https://cloudflare.com
```

## Expected behavior
A readable site is recorded and replayed instead of ��������

## Screenshots
![image](https://user-images.githubusercontent.com/511499/51600020-03a94800-1ecf-11e9-98bd-af26531c360b.png)

## Environment
- OS: macOS 10.14
- Browser: Google Chrome 72.0.3626.64 beta
- Version: Python 3.7.2 and pywb v2.1.0 [e.g. 22]

N0taN3rd:
Have you tried using a non-empty config file and ensured that pywb auto-generated a CA ([docs](https://pywb.readthedocs.io/en/latest/manual/configuring.html#https-proxy-and-pywb-certificate-authority)) and or the dev branch?

I was not able to reproduce the issue using the config below when running from the dev branch using 3.7.2 and linux.
 
```yml
collections:
    all: $all
    live: $live

# Settings for each collection
use_js_obj_proxy: true

# Memento support, enable
enable_memento: true

# Replay content in an iframe
framed_replay: true
```

pirate:
It correctly autogenerated a CA, and the issue persists with the config you gave, except now the frame is tiny in the upper left corner of the screen:

<img width="625" alt="screen shot 2019-01-24 at 4 33 11 pm" src="https://user-images.githubusercontent.com/511499/51709956-d0ada400-1ff5-11e9-9694-a77c09dd9d2d.png">

ikreymer:
[...]

PR #444, opened 2019-02-19
Brotli: Don't accept brotli if library can't be loaded.

ikreymer:
## Description
If `import brotli` fails, print warning and remove `br` from `Accept-Encoding`
Otherwise, can fail silently when trying to record/replay brotli content.

## Motivation and Context

See #434

## Types of changes

## Checklist:

### DIFF
--- a/pywb/rewrite/rewriteinputreq.py
+++ b/pywb/rewrite/rewriteinputreq.py
@@ -5,10 +5,11 @@ from six import iteritems
 from six.moves.urllib.parse import urlsplit
 import re
 
+
 try: # pragma: no cover
     import brotli
     has_brotli = True
-except:  # pragma: no cover
+except Exception:  # pragma: no cover
     has_brotli = False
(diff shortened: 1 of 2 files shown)


pywb/rewrite/rewriteinputreq.py
```diff
@@ -5,6 +5,13 @@
 from six.moves.urllib.parse import urlsplit
 import re
 
+try: # pragma: no cover
+    import brotli
+    has_brotli = True
+except:  # pragma: no cover
```
N0taN3rd: I think it would be good practice to start using named exceptions when possible.

Here it would be `except ImportError` in order to be clearer about why we expect an exception and to be pythonic :)

ikreymer: In general, yes, but here want to catch any `Exception` as that means brotli is not properly installed. Added `Exception`

PR #454, opened 2019-03-10
Support for default timestamp/date for proxy mode

### LINKED ISSUE #452: Possibility to define access date for proxy mode
When accessing a collection via proxy mode, it is not possible to set a desired date.

Proposed solution: offer a command line switch to start pywb with a default access date for proxy mode, for instance `wayback --proxy-dt 200507241832`

This would make pywb usable for creating  software and network environments that are set to a certain time. Concrete context: in the software preservation context, legacy software accesses web resources that aren't available anymore and would be displayed inside the software. The software wouldn't respect system-wide proxy settings and not allow local proxy settings. Web traffic could be routed through pywb via a transparent proxy, providing a curated collection of for example online instruction manuals, plus fallback to public web archives with the default access time set to the release date or usage date of the software.

One pywb instance would be launched per software environment, with only a single collection defined, so the default date time could be set globally.

ikreymer:
## Description

Adds a `--proxy-timestamp` cli flag to specify default timestamp for proxy mode.

Also updates docs, including mention of memento support in proxy mode.

Additional fix: remove `WB_wombat_` from POST append requests for better compatibility with old pywb captures. 

## Motivation and Context

Fixes #452 

POST query 'unrewriting' fixes replay of clicked images from  https://webrecorder.io/despens/amalia-ulman-excellences--perfections collection

## Types of changes

## Checklist:


pywb/apps/cli.py
```diff
@@ -76,7 +78,8 @@ def __init__(self, args=None, default_port=8080, desc=''):
             self.extra_config['proxy'] = {
                 'coll': self.r.proxy,
                 'recording': self.r.proxy_record,
-                'enable_wombat': self.r.proxy_enable_wombat
+                'enable_wombat': self.r.proxy_enable_wombat,
+                'timestamp': self.r.proxy_timestamp,
```
N0taN3rd: Should this be `proxy_timestamp` instead of `timestamp`?
I think this would get confusing as `timestamp` would have to meaning:
- replay datetime
- proxy datetime

N0taN3rd: Its used by frontendapp.py as 
```python3
self.proxy_timestamp = proxy_config.get('timestamp')
```

pywb/apps/rewriterapp.py
```diff
@@ -135,6 +135,10 @@ def _check_accept_dt(self, wb_url, environ):
 
                 wb_url.type = wb_url.REPLAY
 
+            elif 'pywb_proxy_timestamp' in environ:
+                wb_url.timestamp = environ['pywb_proxy_timestamp']
```
N0taN3rd: Likewise re comment about internal naming convention

pywb/apps/frontendapp.py
```diff
@@ -559,6 +564,14 @@ def init_proxy(self, config):
         else:
             self.proxy_prefix = '/{0}/id_/'.format(proxy_coll)
 
+        self.proxy_timestamp = proxy_config.get('timestamp')
+        if self.proxy_timestamp:
+            if not self.ALL_DIGITS.match(self.proxy_timestamp):
+                try:
+                    self.proxy_timestamp = iso_date_to_timestamp(self.proxy_timestamp)
+                except:
```
N0taN3rd: bare except: we should use `except Exception`

docs/manual/configuring.rst
```diff
@@ -429,6 +429,14 @@ To enable proxy mode, the collection can be specified by running: ``wayback --pr
 
   proxy:
     coll: my-coll
+
+The timestamp/date can also be optionally specified by running: ``wayback --proxy my-coll --timestamp 20181226010203`` or by specifying the config::
```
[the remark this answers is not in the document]
ikreymer: It also accepts ISO Date for convenience.. not sure a better way to word it, but yeah, probably makes sense for consistency.

pywb/apps/cli.py
ikreymer: Maybe it should be `default_timestamp`?

The reason is in the yaml config its already in the proxy options
```
proxy:
   coll: ...
   proxy_timestamp:
```
so proxy_timestamp seems redundant..

Naming is hard :)

pywb/apps/frontendapp.py
ikreymer: ok, will update

pywb/apps/cli.py
N0taN3rd: #hardestproblemcompsci

Issue #459, opened 2019-04-10
Recording fails to render


edsu:
I suspect you already know about web content that fails to capture, but in case this one is new: https://www.newblackmaninexile.net/ ... it seems to capture when recording, but if you go to the recording later it fails to render. I tried downloading the WARC file and playing back, and the same thing happens.

ikreymer:
@edsu thanks! Turned out to be a simple fix, I think.. (moved the issue from Webrecorder repo) 
The JSONP rewriter was too strict with comment.. Should be fixed in the associated PR if you want to try.

PR #460, opened 2019-04-10
JSONP Rewriter: Fix regex to match both /* and // comments

ikreymer:
## Description
JSONP Rewriter before only allowed matching function being matched, eg. `/* comment */ jQuery({`
Also support matching JSONP:
```
// Comment
jQuery({
```
as browsers also accept this as JSONP

## Motivation and Context

Fixes replay issue in #459 

https://www.newblackmaninexile.net/

## Types of changes

## Checklist:


ikreymer:
Thanks, fixed tests and supporting multiline comments better

PR #469, opened 2019-05-24
Allow pywb to handle urllib3 >= 1.25.x

### LINKED ISSUE #467: urllib3 version >= 1.25 breaks ssl
It appears that urllib3 has made some breaking changes in versions >= 1.25 that the requests package has not accounted for.

Downgrading urllib3 to version 1.24.3 fixes the issue.

Noting this here per support email.

N0taN3rd:
## Description

Added custom requests HTTPAdapter, PywbHttpAdapter, that restores the behavior of urllib3 < 1.25.x 
which was to not verify ssl certs IFF the installed urllib3 version minor is greater than 24  fixes #467 

Increased the maximum HTTP line length to double the default length in order to ensure super long HTTP lines http are allowed 

## Motivation and Context

urllib3 is now verifying ssl certificates by default.
This is a breaking change from its previous which was to **not verify by default**  

## Types of changes

## Checklist:

### DIFF
--- a/pywb/warcserver/http.py
+++ b/pywb/warcserver/http.py
@@ -1,27 +1,53 @@
-from requests.adapters import HTTPAdapter
-import requests
 import os
 
+import requests
 import six.moves.http_client
+from requests.adapters import DEFAULT_POOLBLOCK, HTTPAdapter
+from urllib3.poolmanager import PoolManager
+
 six.moves.http_client._MAXHEADERS = 10000
+six.moves.http_client._MAXLINE = 131072
 
 SOCKS_PROXIES = None
 orig_getaddrinfo = None
 
 
-#=============================================================================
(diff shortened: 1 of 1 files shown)


pywb/warcserver/http.py
```diff
@@ -1,27 +1,66 @@
...
+
 six.moves.http_client._MAXHEADERS = 10000
+six.moves.http_client._MAXLINE = 131072
 
 SOCKS_PROXIES = None
 orig_getaddrinfo = None
 
 
-#=============================================================================
+def should_use_custom_adaptor():
```
ikreymer: To simplify, think we should just use the pywb adapter always, no need to check version.
The adapter init is simply enough and backwards compatible

N0taN3rd: done

PR #473, opened 2019-05-31
wombat postMessage override tweaking

N0taN3rd:
## Description

removed the definition of `__WB_pmw` from `ensureServerSideInjectsExistOnWindow` in order to allow more proper handling of that definition to occur from `initNewWindowWombat` or `wombatInit`.

`initNewWindowWombat` now initializes wombat for (i)frame's with src values prefixed with `about:` as `about:srcdoc` is commonly used

tweaked postMessage and event listener overrides to be more like the previous wombat revision

## Motivation and Context

URL requiring change: https://twitter.com/webrecorder_io (any twitter profile url will work as well)

It is safer to allow `initNewWindowWombat` to decide if `__WB_pmw` should be defined rather than defining it 
before the code that made the decision to apply the overrides code is executed.

The current behavior was used to handle some edge cases of (i)frame injection where the timing of wombat's initialization does not occur first 
or is skipped because the `src` value was starting some `about:` suffix not currently recognized.
Rather than looking only for `about:blank` initialize wombat manually for (i)frames with any `about:` prefix   

## Types of changes

## Checklist:


wombat/src/wombat.js
```diff
@@ -4736,7 +4730,7 @@ Wombat.prototype.initPostMessageOverride = function($wbwindow) {
       targetOrigin = this_obj.location.origin;
     }
 
-    // console.log("Sending " + from + " -> " + to + " (" + targetOrigin + ") " + message);
```
ikreymer: very minor, but do we want to keep the original syntax to be compatible with ES5, though this is just for debugging and commented out?

N0taN3rd: this line is only for debugging and was changed to accommodate easily adding new things to be logged

PR #480, opened 2019-06-18
Fix #471: failed playback due to encoding issue

### LINKED ISSUE #471: Failed playback due to encoding issue in modern pywb, not in v0.32.0
## In Brief

A website's javascript once made a poorly-formed AJAX POST, which was captured in the [attached warc](https://github.com/webrecorder/pywb/files/3237463/QSQ2-SJ6L.warc.gz). The warc can be played back without problems using pywb 0.32.0, but not with modern pywb (or Webrecorder/ Webrecorder Desktop app)

## At Length

Perma once captured https://www.wired.com/story/our-machines-now-have-knowledge-well-never-understand/, producing [this warc](https://github.com/webrecorder/pywb/files/3237463/QSQ2-SJ6L.warc.gz). The warc contains a peculiar POST request: it claims `content-type: application/x-www-form-urlencoded` in its headers, but the data it posted is not in fact a url-encoded string:
```
# output produced using warcio
# print(record.raw_stream.read())
# print(record.http_headers)

POST /?T=B&u=https%3A%2F%2Fwww.wired.com%2Fstory%2Four-machines-now-have-knowledge-well-never-understand%2F&t=1514932476678&v=1514932476802&Q=1&m=1 HTTP/1.1
origin: https://www.wired.com
content-length: 396
accept-language: en-US,*
accept-encoding: gzip, deflate
host: wired-app.quantummetric.com
accept: */*
user-agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/50.0.2661.37 Safari/537.36
referer: https://www.wired.com/story/our-machines-now-have-knowledge-well-never-understand/
cookie: __cfduid=dd3cdc36c20f0ddb575b25d041573c7671514932471
content-type: application/x-www-form-urlencoded

b'\x816l`L\xa04P\x0e\xe0r\x02\xb5\x89\x19\x00fP\xdb\x0e\xb0\x02,\xee\x0e\x04\x99\xc1\xc5*\x8c\x00\'\xd9\xa1IY\x95\xd0\xb4\xd8\xa6\x94}\x86\x0b@\xc2$\x86\xc9\xb4\x8b\xc0\xbc\x8c\t\x04Wj\xc0\x8a\x1a\xbdB\xf5#\x80\x0fC\t2%a\x1c\x1e\xec\x0c\x05\xa0F\xc2\x12U\xc84\xc8L\x88TX\xb6\x05\xb4"\xf3\xabB\x10\x1aj\xe5\x19\x0b\xa3\xd3!\xc8\x8b\x90@\xe1 `\x84`C\xc2h\x81\x90\xf3i\x11\x11@\xe9p\x0b\xb9\x91\xf0Q@\x81#k\xa0%\xb1a\xb8A\xa5\xd0Xg\x14\xc6$\xab\xd2\xf0H\x8as\xd6\xd3!\x91\xfar\xe2fV\x80!i\x90\x9a\x80\x0bCqXc\xd1t0&5\x91AR\xa6\xb9\x0fd#\xd3\xeaGtL\xe4X\xc6\xa9C\xf0\x0eB\xbb&c\xb0\xc4\x80\xe5\x1ap\n\xc3\x98\x10\xe0!d)h"\xd0\xc0N\x11\x08\xe8\x0b]R\x1c@-C\xc6\x1c\x98\xd4\\\xa9 <\xcds&\xce\xc3\xb1T\xac\x1d8@\x0c\xc2S \xe5n\xbc\x15\xa2\x8b\xce\xdc!\xd8\xe8\xa1O\xcf\x00\x81!\xa0\xa0*\x98\x01G\x86\x0b4\xc9\x04\xb0\xb8\x1aj\x92r\x05\xac3(\xe4\x18\x81\xf7\xcd\xc4d;1+lA\x9e\xc2\xd4\xa0\x90\t\x90\tW\x83\xac\xb1r\xfa\x08N \xaf\x80\xfa\x05\x84\x19x\xe3\x1cz\x87\xc5\xed\x80\xf84\xabb\xe5\x1f5O\x9c\xb9I\x9a\x9c5\xc0!\x89\x80\n\xd2\xa3OI\x93\x82\x93\x98\x10\x15\xae\x16=\x11\x88\xe9\x151\xb6\x1a\x08\xc8\n@c\x188\x0eo\x83\x82B\x11\x80J\xba\x00\x00@'
```
In the process of uploading this warc to Webrecorder, pywb attempts to [convert that data to a native string](https://github.com/webrecorder/pywb/blob/master/pywb/warcserver/inputrequest.py#L232), using the [utility function from warcio](https://github.com/webrecorder/warcio/blob/master/warcio/utils.py#L14), which attempts to decode using 'utf-8'. The decode move fails with a unicode error, of course... which causes upload to fail and playback to fail.

Perma has successfully been playing back this warc since the data of its creation because we've been using version 0.32.0 of pywb, which [decodes as 'iso-8859-1' instead of 'utf-8'](https://github.com/webrecorder/pywb/blob/0.32.0/pywb/utils/loaders.py#L59)...producing gobbledygook output here, but NOT an encoding error.

## Ideas?

I'm not sure what the right way to handle this is. Maybe, similar to [here](https://github.com/webrecorder/pywb/commit/adf34cdb3569ac3f5999cb8b07ff98f8ede168c1), this [whole stanza of code](https://github.com/webrecorder/pywb/blob/master/pywb/warcserver/inputrequest.py#L231-L257) should catch encoding errors, and base64 encode the query if anything goes wrong, like in the [`else`](https://github.com/webrecorder/pywb/blob/master/pywb/warcserver/inputrequest.py#L259)?

I am happy to submit a PR using this strategy, or any strategy you think might be better, for working around the issue.

rebeccacremona:
## Description, Motivation and Context

This is a first pass at resolving #471

The new test intentionally demonstrates that this is a more of a workaround than a real "solution" per se.... what a weird-looking, useless query string... Even though this permits playback to proceed (and restores the behavior of earlier versions of pywb), I wonder if we might be able to do better by base64-encoding it and appending it to __wb_post_data= as in the else clause... but I'm not really qualified to speculate on these topics at this point lol.

Happy to iterate on this / adjust in any way.

## Types of changes

## Checklist:

### DIFF
--- a/pywb/warcserver/inputrequest.py
+++ b/pywb/warcserver/inputrequest.py
@@ -234,6 +234,7 @@ class MethodQueryCanonicalizer(object):
                 query = unquote_plus(query)
             except UnicodeDecodeError:
                 query = to_native_str(query, 'iso-8859-1')
+                query = unquote_plus(query, 'iso-8859-1')
 
         elif mime.startswith('multipart/'):
             env = {'REQUEST_METHOD': 'POST',
--- a/pywb/warcserver/inputrequest.py
+++ b/pywb/warcserver/inputrequest.py
@@ -236,7 +236,7 @@ class MethodQueryCanonicalizer(object):
 
         if mime.startswith('application/x-www-form-urlencoded'):
             try:
-                query = to_native_str(query)
+                query = to_native_str(query.decode())
                 query = unquote_plus(query)
             except UnicodeDecodeError:
                 query = handle_binary(query)
--- a/pywb/warcserver/inputrequest.py
+++ b/pywb/warcserver/inputrequest.py
@@ -229,8 +229,11 @@ class MethodQueryCanonicalizer(object):
             mime = ''
 
         if mime.startswith('application/x-www-form-urlencoded'):
-            query = to_native_str(query)
-            query = unquote_plus(query)
+            try:
+                query = to_native_str(query)
+                query = unquote_plus(query)
+            except UnicodeDecodeError:
+                query = to_native_str(query, 'iso-8859-1')
 
         elif mime.startswith('multipart/'):
             env = {'REQUEST_METHOD': 'POST',
--- a/pywb/warcserver/inputrequest.py
+++ b/pywb/warcserver/inputrequest.py
@@ -236,7 +236,7 @@ class MethodQueryCanonicalizer(object):
 
         if mime.startswith('application/x-www-form-urlencoded'):
             try:
-                query = to_native_str(query.decode())
+                query = to_native_str(query.decode('utf-8'))
                 query = unquote_plus(query)
             except UnicodeDecodeError:
                 query = handle_binary(query)
--- a/pywb/warcserver/inputrequest.py
+++ b/pywb/warcserver/inputrequest.py
@@ -228,13 +228,18 @@ class MethodQueryCanonicalizer(object):
         if not mime:
             mime = ''
 
+        def handle_binary(query):
+            query = base64.b64encode(query)
+            query = to_native_str(query)
+            query = '__wb_post_data=' + query
+            return query
+
         if mime.startswith('application/x-www-form-urlencoded'):
             try:
                 query = to_native_str(query)
                 query = unquote_plus(query)
             except UnicodeDecodeError:
-                query = to_native_str(query, 'iso-8859-1')
-                query = unquote_plus(query, 'iso-8859-1')
+                query = handle_binary(query)
 
         elif mime.startswith('multipart/'):
             env = {'REQUEST_METHOD': 'POST',
@@ -260,9 +265,7 @@ class MethodQueryCanonicalizer(object):
             query = self.amf_parse(query, environ)
 
         else:
-            query = base64.b64encode(query)
-            query = to_native_str(query)
-            query = '__wb_post_data=' + query
+            query = handle_binary(query)
 
         self.query = query
 
--- a/pywb/warcserver/test/test_inputreq.py
+++ b/pywb/warcserver/test/test_inputreq.py
@@ -6,7 +6,7 @@ from six.moves.urllib.parse import parse_qsl
 from io import BytesIO
 from pyamf import AMF3
 from pyamf.remoting import Request, Envelope, encode
-
+from warcio.utils import to_native_str
 
 #=============================================================================
 class InputReqApp(object):
@@ -140,7 +140,7 @@ class TestPostQueryExtract(object):
         mq = MethodQueryCanonicalizer('POST', 'application/x-www-form-urlencoded',
                                 len(data), BytesIO(data))
 
-        assert mq.append_query('http://example.com/') == 'http://example.com/?\x816l`L\xa04P\x0eàr\x02µ\x89\x19\x00fPÛ\x0e°\x02,'
+        assert mq.append_query('http://example.com/') == 'http://example.com/?' + to_native_str(data, 'iso-8859-1')
 
     def test_options(self):
         mq = MethodQueryCanonicalizer('OPTIONS', '', 0, BytesIO())
--- a/pywb/warcserver/test/test_inputreq.py
+++ b/pywb/warcserver/test/test_inputreq.py
@@ -135,6 +135,13 @@ class TestPostQueryExtract(object):
 
         assert mq.append_query('http://example.com/') == 'http://example.com/?foo=bar&dir=/baz'
 
+    def test_post_extract_malformed_form_data(self):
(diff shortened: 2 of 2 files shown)


pywb/warcserver/inputrequest.py
```diff
@@ -229,8 +229,11 @@ def __init__(self, method, mime, length, stream,
             mime = ''
 
         if mime.startswith('application/x-www-form-urlencoded'):
-            query = to_native_str(query)
-            query = unquote_plus(query)
+            try:
+                query = to_native_str(query)
+                query = unquote_plus(query)
+            except UnicodeDecodeError:
+                query = to_native_str(query, 'iso-8859-1')
```
N0taN3rd: do you think that we should also perform an `unquote_plus` on the `iso` decoded query in order to perform the same operations in the `utf-8` case?

rebeccacremona: Hmmm. I'm having trouble reasoning about this, because at least in the single example I have, we're essentially processing junk (some bytes that never were url-encoded form data), right? I left it out thinking this was a lost cause and `unquote_plus`ed junk was no better than the original; on the other hand I can't imagine it'd be any worse! And I suppose if there's any chance at all that `query` might ACTUALLY be `iso-8859-1` url-encoded form data, then totally, performing an `unquote_plus` sounds like the right move.... I'll make that change momentarily.

rebeccacremona: Pushed! Thanks for taking a look!

ikreymer:
Hm, I think that falling through to the else and adding the base64 encoded value to `__wb_post_data=` would to be the better option. The reason for the parsing form-encoded data is to be able to treat the POST params the same way as query params (eg. for fuzzy matching):
For example:

```
POST /path?A=B
Host: example.com/
Content-Type: application/x-www-form-urlencoded

C=D
```

would be added to the index as `example.com/path/?A=B&C=D`

However, if the POST body were to be generic binary data, then we index it as `example.com/path/?A=B&__wb_post_data=<base64 data>`.

It seems that in this case the decoding is failing because it is generic binary data and not form-encoded, it should be treated the same way. According to spec, form-encoded data should generally be UTF-8 anyway https://url.spec.whatwg.org/#concept-urlencoded-serializer, so if its failing, it may be best to just treat it as binary.

rebeccacremona:
Thanks for the guidance! Done.

ikreymer:
Thanks, looks good (just 2.7 test issue remaining)

rebeccacremona:
The horror, the horror: so that of course is a true difference in behavior between python 2 and 3. If we want to detect that this is binary, not `utf-8`, in python 2, we're going to have to do something goofy.

In python 2, of course a `UnicodeError` is not thrown here, since `to_native_str` correctly observers that a byte-string of binary data is already a "native string"... so the original code path is still used, resulting in a url like ```http://example.com/?\x816l`L\xa04P\x0e\xe0r\x02\xb5\x89\x19\x00fP\xdb\x0e\xb0\x02```. In python 3, a UnicodeError is thrown when `to_native_str` converts the byte-string to a unicode-string, and so uses the `except` path, resulting in a url like `http://example.com/?__wb_post_data=gTZsYEygNFAO4HICtYkZAGZQ2w6wAiw=`. 

To get 2/3 parity + `__wb_post_data=<base64 data>` I can make the somewhat wacky move of running `query.decode()` first, forcing the issue. To me, that feels unfortunate but correct... 

Or I can drop 9e2587, going back to using `iso-8859-1` to encode, and not using `__wb_post_data=<base64 data>`. Which seems safer, more like a "non-breaking change"... but less correct...

I pushed a commit adding `query.decode()`, in case your instincts still say `__wb_post_data` is the way to go.

pywb/warcserver/inputrequest.py
```diff
@@ -228,9 +228,18 @@ def __init__(self, method, mime, length, stream,
...
+        def handle_binary(query):
+            query = base64.b64encode(query)
             query = to_native_str(query)
-            query = unquote_plus(query)
+            query = '__wb_post_data=' + query
+            return query
+
+        if mime.startswith('application/x-www-form-urlencoded'):
+            try:
+                query = to_native_str(query.decode())
```
ikreymer:
```suggestion
                query = to_native_str(query.decode('utf-8'))
```
Yep this makes sense! Want to ensure checking for utf-8, so adding it explicitly

rebeccacremona: Righto.

PR #482, opened 2019-06-24
Cap extended query string length when indexing warcs.

rebeccacremona:
## Description
When indexing warcs, pywb parses form-encoded data "so that POST params can be treated the same way as query params (eg. for fuzzy matching)" (see comment from ikreymer in #480). This PR suggests a limit of 32K for the length of the extended query string.

## Motivation and Context
We have a capture of https://www.huffpost.com/entry/the-time-has-come-to-say_b_478388 (https://perma.cc/N3RY-MJ62) that, when played back in a browser without ad blockers, plays back a truly ridiculous POST once made to Facebook's user-tracking API. See https://github.com/harvard-lil/perma/issues/2634 for details. The POST includes a great deal of url-encoded data, which is [appended to the url](https://github.com/rebeccacremona/pywb/blob/develop/pywb/warcserver/inputrequest.py#L183) during indexing.

The resulting gargantuan url is stored in redis without a problem, and is [returned intact by the resource handler/loader](https://github.com/webrecorder/pywb/blob/a2ea925d17fabecd3575c94b41f3923cb19051c8/pywb/warcserver/resource/responseloader.py#L59), but by the time the rewritterapp [tries to process the`Warcserver-Cdx` header](https://github.com/webrecorder/pywb/blob/455efb17ad5f1679dbbf32e5c670c5c9510e5d57/pywb/apps/rewriterapp.py#L309), the CDXJ has been truncated, mid query string, resulting in a `CDXException`.

I did not successfully track down where the truncation is taking place. But, through trial and error, working by powers of two, I determined that a query string of 32K passes through intact while a string of 64K does not. 

It seems a bit of a shame to cap the size of the query string here since we get so far in the process before running into trouble.... but 32K seems more than sufficient to assist with fuzzy matching, and even if we figure out what limit we're running into here (bottle? uwsgi? nginx? docker?), and determine how to raise it, at least to me it feels a little goofy to raise it without some clear benefit.

What do you folks think? Is this a reasonable spot for an intervention?

## Types of changes

## Checklist:


ikreymer:
Thanks for PR, sorry to have messed it up during rebase.
PR #519 now supersedes this, applies the limit earlier, avoids reading a large POST body (>16384) into memory.

PR #483, opened 2019-06-27
Update configuring.rst to reflect default config.yaml.

ekilfeather:
The Docs specify the default value for the warc files path as 'archives' but the default config.yaml file specifies 'archive'
https://github.com/webrecorder/pywb/blob/master/pywb/default_config.yaml#L4


ikreymer:
Thanks!  We're thinking about a better name for this, but good to have the docs be accurate!

PR #491, opened 2019-07-29
Support/Improvements to Domain Cookie Cache

ikreymer:
## Description

To support domain cookies, pywb must cache cookie headers that have a domain property, eg. `Domain= .example.com`
This was only supported in WR, now also supported in pywb for live and record mode.
The cache is per-collection, so recording/live mode can only be used by a single user in pywb.
The cache is stored in fakeredis by default.
(To support per user, would require pywb to store its own session cookie. In WR, the existing session cookie is used).

Additional improvements to domain cookie cache:
- Don't update cache if response is a service-worker and response is non-200, as browser will not load.
- Don't add duplicate cookies to Cookie and Set-Cookie headers.
- Add `sw_` and `wkrf_` to cookie prefix rewrites.

## Motivation and Context

Use of domain cookies is required to be able to log in to many sites, such as Twitter.

## Types of changes

## Checklist:


pywb/apps/rewriterapp.py
```diff
@@ -366,7 +376,12 @@ def render_content(self, wb_url, kwargs, environ):
                                                    config=self.config))
 
         cookie_rewriter = None
-        if self.cookie_tracker:
+        if self.cookie_tracker and cookie_key:
```
N0taN3rd: Running code inspection on this file reports that cookie_key maybe referenced before assignment here

I would feel better if we assign `cookie_key = None` after line 275 as is done with the variable `setcookie_header`

PR #504, opened 2019-08-29
SOCKS Proxy Improvements

ikreymer:
## Description
Improves the SOCKS proxy handling:
- https urls over SOCKS were sometimes improperly sent using full url, using higher-level adapter to avoid this
- ensure SOCKS proxy uses same cert verification mode as non proxy (currently no checking).
- removed unnecessary monkey patching, as requests/urllib3 now support socks5 without patching.
- also add a way to toggle proxy with env settings SOCKS_DISABLE
- replace lower-level `adapter.get_connection()` with `adapter.proxy_manager_for(self.socks_proxy)` which returns a proxy manager which determines appropriate url to use with socks proxy.

## Motivation and Context

Loading https://check.torproject.org/ over SOCKS resulted in a 500 error, because the proxy was incorrectly sending `GET https://check.torproject.org/` instead of `GET /`.

## Types of changes

## Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,10 +1,14 @@
 pywb 2.3.5 changelist
 ~~~~~~~~~~~~~~~~~~~~~
 
-* General auto-fetch fixes
+* General auto-fetch fixes (#503)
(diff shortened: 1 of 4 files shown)


ikreymer:
Will clarify in commit message

PR #509, opened 2019-10-09
Docs, custom metadata improvements

ikreymer:
## Description

Ensure user-metadata, available in metadata.yaml files is consistently passed to
UI templates, separate from config information.
Pass metadata, if any, to frame insert and don't mix with config (ukwa/ukwa-pywb#57)
Also add docs about how to configure custom metadata.
Update README for 2.4 release

## Motivation and Context

Issue raised in ukwa/ukwa-pywb#57 about mixing user metadata and config,
question about metadata in frame_insert.html

## Types of changes

## Checklist:

### DIFF
--- a/pywb/apps/frontendapp.py
+++ b/pywb/apps/frontendapp.py
@@ -293,7 +293,7 @@ class FrontEndApp(object):
         """Retrieve the collection config, including metadata, associated with a collection
 
         :param str coll: The name of the collection to receive config info for
-        :return: The collections config. The 'metadata' dict contains optional user metadata
+        :return: The collections config
         :rtype: dict
         """
         coll_config = {'coll': coll,
--- a/pywb/apps/frontendapp.py
+++ b/pywb/apps/frontendapp.py
@@ -289,25 +289,22 @@ class FrontEndApp(object):
         except Exception:
             self.raise_not_found(environ, 'static_file_not_found', filepath)
 
-    def get_metadata(self, coll):
-        """Retrieve the metadata associated with a collection
+    def get_coll_config(self, coll):
+        """Retrieve the collection config, including metadata, associated with a collection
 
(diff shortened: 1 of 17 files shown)


pywb/apps/frontendapp.py
```diff
@@ -289,25 +289,22 @@ def serve_static(self, environ, coll='', filepath=''):
...
 
-    def get_metadata(self, coll):
-        """Retrieve the metadata associated with a collection
+    def get_coll_config(self, coll):
+        """Retrieve the collection config, including metadata, associated with a collection
 
-        :param str coll: The name of the collection to receive metadata for
-        :return: The collections metadata if it exists
+        :param str coll: The name of the collection to receive config info for
+        :return: The collections config. The 'metadata' dict contains optional user metadata
```
N0taN3rd:
```suggestion
        :return: The collections config
```

I think that `The 'metadata' dict contains optional user metadata` would work better in the description of the method not the return value. 

At least to me it reads as if a tuple of two dictionaries was being returned

pywb/warcserver/index/fuzzymatcher.py
```diff
@@ -194,6 +194,11 @@ def match_general_fuzzy_query(self, url, urlkey, cdx, rx_cache):
         check_query = False
         url_no_query, ext = self.get_ext(url)
 
+        # don't fuzzy match to 204
+        if cdx.get('status') == '204':
+            if '__pywb_method=options' in cdx['urlkey']:
```
N0taN3rd: do not fuzzy much if 204 and method options?

ikreymer: Yes, if we encounter an OPTIONS response with 204, don't include it as a match.
Mostly for warcprox WARCs which write OPTIONS requests with 204.

PR #536, opened 2020-02-10
Try to remove headers with illegal characters.

danielbicho:
## Description
Try to remove headers with illegal characters. Fixing the following issue arquivo/pwa-technologies#774

## Motivation and Context
Pywb is not able replay records with illegal characters (outside latin1) in the replay headers.
In example above there is a Trademark character on the Server header that breaks the streaming of the response.

This change try to fix those badly formed headers, detecting them and removing them. Its a best enforce approach, in last case the record will fail to load as it was.

arquivo/pwa-technologies#774

image

After this fix:
image

## Types of changes

I am not sure if this is the best approach, open to do any change needed to make it go upstream.

### DIFF
--- a/pywb/apps/wbrequestresponse.py
+++ b/pywb/apps/wbrequestresponse.py
@@ -142,6 +142,18 @@ class WbResponse(object):
         response.add_access_control_headers(env=env)
         return response
 
+    def try_fix_errors(self):
+        """Utility method to try remove faulty headers from response.
+
+        :return:
+        :rtype: None
+        """
+        for header in self.status_headers.headers:
+            try:
+                header[1].encode('latin1')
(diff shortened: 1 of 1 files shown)


ikreymer:
Thanks! I think this makes sense, though do you know what's causing the invalid headers? Is it UTF-8 encoding, just wondering?

danielbicho:
Yes it is utf8 encoding.

Opening the WARC Record with VIM we can see the problematic line (My vim is assuming latin1 encoding):
Server: Casper Cacheâ<84>¢  V2 (CAST TIME)^M

Everything goes well, from reading the WARC record to the building of the StatusAndHeaders, but then the WSGI start_response call breaks.

PR #557, opened 2020-03-28
proxy mode: respect settings when started from cli

micronn:
If you configure some proxy settings like enable_content_rewrite in config.yaml, when you run wayback --proxy mycollection it does not work.

This is because the entire proxy section is replaced by the extra config (custom_config) generated according to the command line arguments in cli.py.

This fix makes the section values to be replaced, but not the entire section. So the existing values won't be removed.

Another approach would be to add command line arguments for the rest of the available settings  like --proxy-enable-content-rewrite, etc... I don't know which is preferred in this project.


ikreymer:
Thanks, this is reasonable workaround for now.

Issue #558, opened 2020-04-22
Installation from sdist fails


jayvdb:
## Describe the bug
setup.py fails because requirements.txt is not included in the sdist

## Steps to reproduce the bug
1. Fetch and unpack sdist from PyPI
2. Run setup.py bdist_wheel (or similar)

## Expected behavior
setup.py should run setuptools

PR #559, opened 2020-04-22
Fix sdist

jayvdb:
Fixes https://github.com/webrecorder/pywb/issues/558


jayvdb:
MANIFEST.in generated using `check-manifest` and slightly tweaked to be a bit more descriptive.

ikreymer:
Thanks for updating!

Issue #585, opened 2020-10-11
OutbackCDX does not get parameters of POST request


kaij:
## Describe the bug
When using OutbackCDX as an index server, the __wb_post_data is not sent with the url to the outbackcdx server. On webpages with multiple XHR POSTs to the same URL, this will return the wrong data. Using a local CDXJ file index works as expected.

## Steps to reproduce the bug
1. Create warc of public page at http://www.corona-data.ch/.
2. Index with outbackcdx using a command similar to ```cdx-indexer -p -s corona-data.warc.gz  | curl -X POST --data-binary @- http://127.0.0.1:8078/collection```
3. Open the page in pywb replay -> most of the diagrams will stay white. 

## Expected behavior
The replayed POST requests should contain correct responses (so the diagrams can be drawn)

## Screenshots
Replayed page with invalid (white) diagrams. The reason for this is that the CDX information for the POST requests to _dash-update-components are not passed with the query.
![image](https://user-images.githubusercontent.com/1143418/95678375-940c4c80-0bcc-11eb-8c2b-31ceae1f0746.png)
![image](https://user-images.githubusercontent.com/1143418/95678352-763ee780-0bcc-11eb-9ff0-a62265d3c370.png)

## Environment
- pywb 2.4.2

## Additional context
I tried to track this down to the `_get_api_url` function in warcserver/indexsource.py. The url used does not contain the `__wb_post_data`. `FileIndexSource` uses the `key` parameter. So I see the following options:
- Passing the `key` using the `urlkey` parameter of outbackcdx (and updating documentation)
- Adding `__wb_post_data` to the url parameter

There might be also be other options to consider. Also, the __wb_post_data changed to __warc_post_data with cdxj-indexer, so maybe there is more development going on. I'd be interested to contribute a fix, but need some guidance as to the best way.

ato:
OutbackCDX does not currently have any support for indexing POST responses (pull requests welcome though).

kaij:
@ato I added a first pull request at https://github.com/nla/outbackcdx/pull/91. It would be great if you could have a look at it - feedback and discussion welcome. 😀

ikreymer:
Sorry for not responding earlier!

Most of the POST matching is done on form data (application/x-www-form-urlencoded), then the query params can be matched similar to GET query params.

[...]

PR #587, opened 2020-10-30
Add url_post template parameter for remote cdx api

kaij:
## Description

This change provides a new parameter `url_post` to the template used to configure a remote index source. The `url_post` parameter includes the original url with POST data appended as, if available. The parameter `__wb_post_data` is extracted from the final url key.

## Motivation and Context

This resolves a problem with communicating POST parameters to OutbackCDX. It fixes a replay issue described in #585. OutbackCDX has in the meantime been extended to index POST parameters as described in nla/outbackcdx#91. A sample config.yaml for OutbackCDX looks like this:

```
api_url: http://localhost:9596/nb-webarchive?closest={closest}&sort=closest&url={url_post}
```

I first tried using the existing `alt_url`. However, the alt_url is based on the original (unfiltered) url and it is not available for all calls. In order not to break existing solutions, I propose to use this new parameter.

## Types of changes

## Checklist:

Still missing: tests (work in progress), documentation updates. Feedback and discussion welcome!

### DIFF
--- a/pywb/warcserver/index/indexsource.py
+++ b/pywb/warcserver/index/indexsource.py
@@ -1,4 +1,4 @@
-from six.moves.urllib.parse import quote_plus
+from six.moves.urllib.parse import quote_plus, quote, parse_qs, urlparse
 from warcio.timeutils import PAD_14_DOWN, http_date_to_timestamp, pad_timestamp, timestamp_now, timestamp_to_http_date
 
 from pywb.utils.binsearch import iter_range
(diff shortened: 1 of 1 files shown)


ikreymer:
Thanks for adding this! Since OutbackCDX now supports `__wb_post_data`, I wonder if it would make sense to simply use that parameter to keep it simpler to maintain? For example, maybe JSON data could be prefixed with `__wb_post_data=json:...`?

And yes, would be very helpful to have some tests, let me know if you need any questions on where to add them!

PR #592, opened 2020-11-16
Fix warc pattern

PsypherPunk:


PsypherPunk:
Apologies, didn't correctly check the target.

Issue #598, opened 2020-12-04
Indexing errors on Content-Type: multipart/form-data when "boundary" is missing


ldko:
## Describe the bug

When indexing a WARC file with records containing `Content-Type: multipart/form-data` (missing "boundary" such as in `multipart/form-data; boundary=----WebKitFormBoundaryrdRXu11VSoXKFFBV`), the indexing fails at:

https://github.com/webrecorder/pywb/blob/7b51101b040628ce6ceddb7bd79440b03c0081d4/pywb/warcserver/inputrequest.py#L262

with `ValueError: Invalid boundary in multipart form: b''`

## Steps to reproduce the bug
Download this [sample WARC](https://webarchive.library.unt.edu/thumbs/UNT-20201119192756453-00043-oirxa9z4.warc.gz) (created with Brozzler) that contains records with `Content-Type: multipart/form-data`.
Try to index the WARC.

## Expected behavior
The indexing process should not choke on the inadequate Content-Type header.

## Environment

- OS: Ubuntu 18.04.5
- Python 3.7.6

ldko:
Fixed in #599 .

PR #599, opened 2020-12-04
Handle Content-Type multipart/form-data without boundary

ldko:
## Description

Catches `ValueError` on a multipart/form-data mime when "boundary" is missing, and handles it in the same way as an earlier condition for 'application/x-www-form-urlencoded' erroring on `UnicodeDecodeError`.  When I try to run the tests with `python setup.py test`, test_proxy.py tests fail, but this seems unrelated and is the same when I run them on the `master` branch.

## Motivation and Context

Fix for #598 to handle a `ValueError` on a multipart/form-data mime when "boundary" is missing.

## Types of changes

## Checklist:

### DIFF
--- a/pywb/indexer/test/test_indexing.py
+++ b/pywb/indexer/test/test_indexing.py
@@ -463,6 +463,104 @@ com,example)/xyz.pdf 20140401052011 http://example.com/xyz.pdf application/http 
 """
 
 
+def test_multipart_form():
+    test_data = b'\
+WARC/1.0\r\n\
+WARC-Type: response\r\n\
+WARC-Record-ID: <urn:uuid:073fac44-c383-4a2b-980d-76fec83bd20d>\r\n\
+WARC-Date: 2020-11-19T19:54:34Z\r\n\
+WARC-Target-URI: https://example.com/ajax/bz?foo=bar\r\n\
+Content-Type: application/http;msgtype=response\r\n\
+WARC-Payload-Digest: sha1:3I42H3S6NNFQ2MSVX7XZKYAYSCX5QBYJ\r\n\
+Content-Length: 48\r\n\
+WARC-Block-Digest: sha1:XN45YTSBLG5PLJ4HA7DRDYGJBM5VW4UO\r\n\
+\r\n\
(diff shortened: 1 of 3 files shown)


ikreymer:
Out of curiosity, what version of python are you running with?
I have not been able to repro this error with the supplied WARC in Python 3.6+ in either Ubuntu or OSX, wonder if it only happens in an older python? But, good to have just in case anyway.

ldko:
Hrrrm...I have gotten the error on Ubuntu 18.04.5 running in a Python 3.7.6 env and a 3.6.9 env when trying to index [this  warc](https://webarchive.library.unt.edu/thumbs/UNT-20201119192756453-00043-oirxa9z4.warc.gz) and another (autoindex or using `wb-manager index`). Running `python setup.py install` on that machine in Python 3.4 or 3.5, the install fails with trying to install the markupsafe depdendency, but I haven't looked into that.

I have also tried this on another machine running Ubuntu 20.04 in Python 3.8.5 and see the same issue:
```
$ wb-manager init test
$ wb-manager add test UNT-20201119192756453-00043-oirxa9z4.warc.gz 
2020-12-05 10:49:21,350: [INFO]: Copied /home/gitty/pywb/UNT-20201119192756453-00043-oirxa9z4.warc.gz to /home/gitty/pywb/collections/test/archive
Error: Invalid boundary in multipart form: b''
```

ikreymer:
Ah, thanks for the steps. Heh, my mistake, i was testing it with just cdx-indexer which of course did not have the write settings by default!

ikreymer:
Could you add a quick test record with an invalid boundary?
Probably simplest would be to add to: https://github.com/webrecorder/pywb/blob/master/pywb/indexer/test/test_indexing.py similar to the other examples there.

ldko:
@ikreymer I have added a few tests: one for a regular well-formed multipart/form-data type that passes in master and this branch and a couple tests where the boundary part is missing which fail in master and pass on this branch. Let me know if you want to see any changes. Thanks!

ikreymer:
Thanks!

PR #603, opened 2020-12-29
Catch uWSGI TypeError for invalid headers

kaij:
Catch TypeError thrown by uWSGI in addition to UnicodeError

## Motivation and Context

The fix for invalid headers merged with webrecorder/pywb#536 seems not to be working with Docker/uWSGI/Python3. Looking at https://github.com/unbit/uwsgi/blob/master/plugins/python/wsgi_headers.c#L140, uWSGI is throwing a TypeError, not a UnicodeError. However, UnicodeError is correct when not using uWSGI (e.g. run with cli.py).

Catching the TypeError solves the replay problem also when running pywb with Docker/uWSGI/Python3.

## Types of changes

## Checklist:

### DIFF
--- a/pywb/apps/wbrequestresponse.py
+++ b/pywb/apps/wbrequestresponse.py
@@ -164,7 +164,7 @@ class WbResponse(object):
         try:
             start_response(self.status_headers.statusline,
                            self.status_headers.headers)
-        except UnicodeError:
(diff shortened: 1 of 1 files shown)


ikreymer:
That makes sense, thanks for the fix!

PR #604, opened 2021-01-05
Fix default banner css namespacing

kaij:
## Description
Added prefix/namespace to css class "mobile" (new: "_wb_mobile") in the default files. 

## Motivation and Context
As the banner CSS may get injected, this can lead to conflicts with the replayed site if a "mobile" class is used.

## Types of changes

## Checklist:

### DIFF
--- a/pywb/static/default_banner.css
+++ b/pywb/static/default_banner.css
@@ -159,7 +159,7 @@
     overflow: scroll;
 }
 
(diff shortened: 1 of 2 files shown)


ikreymer:
Ah yes, this was intended for use with the framed replay, but can be an issue in frameless mode. Nice find!

PR #606, opened 2021-01-20
Default closest_limit to 100 instead of 10

anjackson:
## Description

Raise the default limit of the number of records returned when querying CDX for the closest records.

## Motivation and Context

At UKWA we're hitting cases where crawl variation means we have e.g. a lot of redirect records and in these cases the 10 record limit is too low.  I can't see any way of configuring this value, so I'm proposing the default is raised.

## Types of changes

## Checklist:


ikreymer:
Thanks! Yeah, I think that should be fine, I don't know if there's a need to set the limit lower, so this seems sufficient.
Can add an option to configure the limit if it becomes an issue.

PR #620, opened 2021-03-05
Fix ACL loading for auto collections

sebastian-nagel:
To make the ACL checker work for automatic collections: pass the collection name to the ACL checker, so that the ACL loader is able to expand the template paths (`{coll}/acl/`).


ikreymer:
Thanks for the fix, checked locally.. Will try to add some tests, thought this was working before, but clearly not!

PR #631, opened 2021-03-22
Zipnum index: fail if counting pages with filter params

sebastian-nagel:
## Description

Do not apply any filters (param filter, from, to, closest) if counting pages (param showNumPages=true) to avoid that the CDX filter/processing fails if it tries to read keys not present in a page-count result.

## Motivation and Context

Avoid error responses "HTTP 500 Internal Error: 'urlkey'" (also 'url' or 'timestamp') which are not helpful.

Ev., could also argue to turn these into bad request responses. The solution here tries to be backward-compatible to PyWB 0.33.2 where filter params are simply ignored.

## Types of changes

## Checklist:

### DIFF
--- a/pywb/warcserver/index/aggregator.py
+++ b/pywb/warcserver/index/aggregator.py
@@ -37,7 +37,9 @@ class BaseAggregator(object):
 
         cdx_iter, errs = self.load_index(query.params)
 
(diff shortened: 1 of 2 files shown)


ikreymer:
Makes sense. The pageCount never did take into account any other filters, so probably ok to ignore to be compatible.

PR #633, opened 2021-04-13
xmlquery: use compressed length when available

ato:
## Description

Use the compressed length when it's available in XmlQueryIndexSource.

The field is confusingly misnamed `compressedendoffset` in the XML but [OpenWayback](https://github.com/iipc/openwayback/blob/c7fac11479ee4447a71ee337f91018f7808757a2/wayback-core/src/main/java/org/archive/wayback/resourceindex/cdx/format/CompressedLengthCDXField.java#L33) and consequently [OutbackCDX](https://github.com/nla/outbackcdx/blob/5094be4e12f76d083edc72ba2af2d8150dc5170b/src/outbackcdx/XmlQuery.java#L151) actually use this for the "S" CDX field (compressed length).

## Motivation and Context

Without this field when WARC files are accessed over HTTP pywb will make open byte range requests which results in a lot more data being read from disk than necessary.

CC @anjackson 

N/A

## Types of changes

## Checklist:
- [ ] All new and existing tests passed.  _(test_force_https failures seem unrelated. I get the same failures with the master branch prior to this change.)_

### DIFF
--- a/pywb/warcserver/index/indexsource.py
+++ b/pywb/warcserver/index/indexsource.py
@@ -314,6 +314,11 @@ class XmlQueryIndexSource(BaseIndexSource):
         cdx['digest'] = self.gettext(item, 'digest')
         cdx['offset'] = self.gettext(item, 'compressedoffset')
         cdx['filename'] = self.gettext(item, 'file')
+
+        length = self.gettext(item, 'compressedendoffset')
+        if length:
+            cdx['length'] = length
+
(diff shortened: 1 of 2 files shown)


ikreymer:
Makes sense, I didn't know this field existed in XmlQueryIndexSource. As long as its safe to assume that `compressedendoffset` is always the length, and *not* the end offset.

PR #655, opened 2021-07-06
fix errors in docs/manual/rewriter.rst

yume-chan:
1. Missing a line break.
    image
2. Thoese Python modules are called [`pywb.rewrite.*`](https://github.com/webrecorder/pywb/blob/81308780ecc17a393b684d320d80dada309b4694/pywb/rewrite/default_rewriter.py), not `pywb.rewriter`


ikreymer:
Thanks!

Issue #656, opened 2021-07-06
Searching for urls with semicolons fails because URL sent to backend is cutoff before the first semicolon in the url


maeb:
https://github.com/webrecorder/pywb/blob/81308780ecc17a393b684d320d80dada309b4694/pywb/static/search.js#L125

The url search parameter (every search parameter) should be encoded like this:
`var query = [window.wb_prefix + '*?url=' + encodeURIComponent(url)]; `

**Example:**

`cdx?url=https://akamai.vgc.no/drfront/images/2021/07/05/c=0,58,2000,440;w=1080;h=238;617673.jpg&output=json`

gets converted in the backend to a request like the following to the cdxj api server:

```
index?url=https%3A//akamai.vgc.no/drfront/images/2021/07/05/c%3D0%2C58%2C2000%2C440&closest=&matchType=prefix
```

EDIT: add example

maeb:
Referencing https://datatracker.ietf.org/doc/html/rfc3986#appendix-A that confirms semicolon is member of the sub-delims set of characters which in turn is allowed in the path segment.

PR #657, opened 2021-07-06
Encode url search parameter when performing query

maeb:
## Description

Searching for URL's with semicolons fails because the url search paramater is not encoded in the request from the frontend to the backend. The backend receives the full URL but in the request to the CDXJ server the request is cutoff at (and including) the first semicolon.
## Motivation and Context

#656 
## Types of changes

- [ X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/pywb/static/search.js
+++ b/pywb/static/search.js
@@ -122,7 +122,7 @@ function clearFilters(event) {
 }
 
 function performQuery(url) {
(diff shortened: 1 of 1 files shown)


ikreymer:
Thanks!

PR #659, opened 2021-07-12
Encode load url properly

### LINKED ISSUE #658: Proper encoding of load_url
Formatting of _load_url_ does not encode the _url_ parameter properly if it ends up in the query string of the configured _url_field_ (_replay_url_):
 
https://github.com/webrecorder/pywb/blob/843fe28ed8cc497c3a11345243dbcfc288455337/pywb/warcserver/index/indexsource.py#L160-L162

Some url's does not survive query parameter parsing unscaded when the _url_ parameter is part of the query string of the _load_url_.

This seems to fix the issue:

```
        cdx[self.url_field] = res_template(self.replay_url, dict(url=cdx['url'],
                                                                 timestamp=cdx['timestamp'],
                                                                 src_coll=source_coll))
```

I believe this is a proper fix without breaking changes, but I am not sure. Shall I post a PR?

maeb:
## Description

URL is not properly encoded in request to warcserver.

## Motivation and Context

Fixes #658.

## Types of changes

Not sure if this is a breaking change or the proper way to fix this.

## Checklist:

### DIFF
--- a/pywb/warcserver/index/indexsource.py
+++ b/pywb/warcserver/index/indexsource.py
@@ -157,9 +157,9 @@ class RemoteIndexSource(BaseIndexSource):
         if name:
             source_coll = params.get('param.' + name + '.src_coll', '')
 
-        cdx[self.url_field] = self.replay_url.format(url=cdx['url'],
+        cdx[self.url_field] = res_template(self.replay_url, dict(url=cdx['url'],
                                                      timestamp=cdx['timestamp'],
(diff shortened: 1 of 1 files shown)


maeb:
I erroneously made this PR :) Was supposed to test it in another repository first. Will make changes to the commits as they contain unintended formatting changes.

ikreymer:
Thanks, yes, I think this should work as `res_template` handles url-encoding also.

Issue #663, opened 2021-07-22
eval rewriting breaks JS that defines object fields named 'eval'


ato:
## Describe the bug

The JavaScript rewrite rules break code that uses "eval" as the name of an object field or method. The [less.js 1.3.0](https://github.com/less/less.js/blob/v1.3.0/lib/less/tree/rule.js#L39) library uses this pattern frequently as it defines many functions named 'eval'  to evaluate the parsed less syntax tree using a pattern like this:

```
      tree.Assignment.prototype = {
        eval: function (a) {
          ...
        }
      }
```

## Steps to reproduce the bug
```
$ cat >test.html <<EOF
<script>
obj = {
  eval: function() { return "OK"; }
};
document.write(obj.eval());
</script>
EOF
$ warcit -o http://example.org/ test.htm
$ wb-manager add test test.html.warc.gz
$ pywb
$ firefox http://localhost:8080/test/2/http://example.org/test.html
```
Correct output: `OK`
Incorrect output: (blank)

An in the wild example of a web page using less-1.3.0.min.js is <https://www.pipenetworks.com/>. Under pywb the page loads without any styles. With the eval rewrites in `regex_rewriters.py` commented out the page loads correctly.

## Expected behavior

While fixing all the eval false positives is probably impractical we could adjust the regex with lookahead/lookbehind assertions to prevent matching `eval:` and `obj.eval()`. That'll of course introduce a false-negative `(true?eval:null)("1+2")` but a construction like that is pretty contrived and hopefully very unlikely to appear in real code.

## Screenshots
N/A

## Environment
pywb 2.5.0
Firefox
Linux

ikreymer:
The pywb 2.6 release (via #668) should fix, as much as possible w/o full parsing.

PR #664, opened 2021-07-22
rewrite: avoid rewriting fields named 'eval' on arbitrary objects

ato:
## Description

This changes the eval rewriting regex rules to not match fields named eval on arbitrary objects like:
```js
obj = { eval: function() { return 1; } };
obj.eval();
```
It also adds a second set of rules to ensure `window.eval` continues to match in cases such as:

```js
x = window.eval;
window.eval(a);
```
Note: This does introduce false-negatives like `obj = window; obj.eval(a)` and `(true?eval:null)(a)`. These seem like they should be rare in normal code though.

## Motivation and Context

Fixes issue #663 in which the rewriter breaks less.js as it uses a lot of objects that have methods named eval that have nothing to do with JavaScript's global `eval()` function.

## Types of changes

## Checklist:

### DIFF
--- a/pywb/rewrite/regex_rewriters.py
+++ b/pywb/rewrite/regex_rewriters.py
@@ -101,9 +101,13 @@ if (!self.__WB_pmw) {{ self.__WB_pmw = function(obj) {{ this.__WB_source = obj; 
 
         rules = [
             # rewriting 'eval(....)' - invocation
-            (r'(?<![$])\beval\s*\(', self.add_prefix('WB_wombat_runEval(function _____evalIsEvil(_______eval_arg$$) { return eval(_______eval_arg$$); }.bind(this)).'), 0),
+            (r'(?<![$.])\beval\s*\(', self.add_prefix('WB_wombat_runEval(function _____evalIsEvil(_______eval_arg$$) { return eval(_______eval_arg$$); }.bind(this)).'), 0),
             # rewriting 'x = eval' - no invocation
-            (r'(?<![$])\beval\b', self.add_prefix('WB_wombat_'), 0),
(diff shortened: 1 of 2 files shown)


ikreymer:
I think the false negatives are definitely a concern, and will likely break other sites, unfortunately.

I think it may be possible to solve this by removing the need for `WB_wombat_eval` altogether.

The idea is that the non-invocation eval is instead rewritten to `self.eval`, which is treated the same as non-global eval invocation. The object proxy in wombat will then add the rewriting.

Normally, the following are equivalent:
```
a = eval
a(...)

and

a = self.eval
a(...)
```

But the latter can be intercepted by a Proxy, if `self` is a Proxy object.

Additionally, in the invocation rewriting, wombat will 'pass through' to the original object if it has a non-native eval.

This work is currently here, will open new PRs for that:
https://github.com/webrecorder/pywb/compare/eval-fix?expand=1
https://github.com/webrecorder/wombat/compare/eval-fix?expand=1

This solves the example you've shared, but probably needs a bit more testing

ato:
Seems promising. I guess one situation where rewriting to `self.eval` might break is if the surrounding code uses the [`self = this` pattern](https://gist.github.com/jashmenn/b306add36d3e6f0f6483). I don't have a good feel for how often that would occur in practice.

ikreymer:
> Seems promising. I guess one situation where rewriting to `self.eval` might break is if the surrounding code uses the [`self = this` pattern](https://gist.github.com/jashmenn/b306add36d3e6f0f6483). I don't have a good feel for how often that would occur in practice.

Ah, good point. There will be a way around that, by declaring another top-level object eg. `let ___wb_eval = self.eval` and using that instead. May be more cumbersome, so lets see if this actually occurs in the wild.

Closing this in favor of #668 for now.

PR #690, opened 2022-01-19
fix: updates wombat.js to the version mentioned in the 2.6.3 release notes

omgoo:
## Description
Updates wombat.js with build-wombat.sh. The static version seemed to be from the last release.


ikreymer:
Thanks for pointing this out! I just fixed this with a separate commit, to avoid the messy commit list here, also bumping to 2.6.4

PR #712, opened 2022-04-25
Encoding image 'srcset'  value including the intrinsic width

yasarkunduz:
## Description
When the image attribute **[srcset](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images)** value contains an unsupported character(https://www.rfc-editor.org/rfc/rfc3986#section-2) then the url will be encoded. In this case the value contains a space and the image intrinsic width which will cause a broken image after encoding.

## Motivation and Context
This resolves the problem with encoding image url excluding the intrinsic width. 

## Types of changes
- [x ] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [ x] I have added or updated tests to cover my changes.
- [ x] All new and existing tests passed.

### DIFF
--- a/pywb/rewrite/html_rewriter.py
+++ b/pywb/rewrite/html_rewriter.py
@@ -177,7 +177,7 @@ class HTMLRewriterMixin(StreamingRewriter):
             return ''
 
         values = (url.strip() for url in re.split(self.SRCSET_REGEX, value) if url)
-        values = [self._rewrite_url(v.strip()) for v in values]
+        values = [self._rewrite_url(v.split(' ')[0].strip()) + (' ' + ' '.join(v.split(' ')[1:])).rstrip() for v in values if v]
         return ', '.join(values)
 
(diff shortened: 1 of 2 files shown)


pywb/rewrite/html_rewriter.py
```diff
@@ -177,7 +177,7 @@ def _rewrite_srcset(self, value, mod=''):
             return ''
 
         values = (url.strip() for url in re.split(self.SRCSET_REGEX, value) if url)
-        values = [self._rewrite_url(v.strip()) for v in values]
+        values = [self._rewrite_url(v.split(' ')[0].strip()) + (' ' + ' '.join(v.split(' ')[1:])).rstrip() for v in values if v]
```
edsu, line 180: It would be nice to see a comment here for what is going on here, and why it is needed.

PR #738, opened 2022-07-14
Ensure CDX status from JSON is a string

edsu:
## Description

If a CDX entry has a status that is an integer instead of a string (which is what ArchiveWeb.page currently generates) will blow up:

```
Traceback (most recent call last):
  File "/Users/edsummers/Projects/pywb/pywb/warcserver/basewarcserver.py", line 77, in __call__
    result = endpoint(environ, **args)
  File "/Users/edsummers/Projects/pywb/pywb/warcserver/basewarcserver.py", line 46, in post_fullrequest
    return handler(params)
  File "/Users/edsummers/Projects/pywb/pywb/warcserver/handlers.py", line 160, in __call__
    out_headers, resp = loader(cdx, params)
  File "/Users/edsummers/Projects/pywb/pywb/warcserver/resource/responseloader.py", line 37, in __call__
    entry = self.load_resource(cdx, params)
  File "/Users/edsummers/Projects/pywb/pywb/warcserver/resource/responseloader.py", line 214, in load_resource
    if not status or not status.startswith(('2', '4', '5')):
AttributeError: 'int' object has no attribute 'startswith'
```

This commit will guard against that by casting status values to a string if they are an integer.

## Motivation and Context

It would be useful to be able to unpack a WACZ file created with ArchiveWeb.page (or other tools) and have it playable in pywb by moving the WARC and CDXJ files into place. Currently this is not possible because WACZ uses an int for the CDXJ status, whereas pywb expects it to be a string in multiple places. In the spirit of the robustness principle this change will be lenient in what is accepted in CDXJ files by casting an int status value to a str.

## Types of changes

- [X] Replay fix (fixes a replay specific issue)

## Checklist:

### DIFF
--- a/pywb/warcserver/index/cdxobject.py
+++ b/pywb/warcserver/index/cdxobject.py
@@ -251,7 +251,11 @@ class CDXObject(OrderedDict):
 
     @classmethod
     def json_decode(cls, string):
-        return json_decode(string, object_pairs_hook=OrderedDict)
+        cdx_block = json_decode(string, object_pairs_hook=OrderedDict)
+        # other parts of pywb expect status to be a string and not an integer
+        if cdx_block and type(cdx_block['status']) == int:
+            cdx_block['status'] = str(cdx_block['status'])
+        return cdx_block
 
 
 #=================================================================


edsu:
Oops, I forgot to use a branch when submitting this PR, so I will resend.

PR #741, opened 2022-07-19
Update test docs

edsu:
## Description

Running the tests with `python setup.py test` does not work due to various version mismatches. However they do run with tox, as they are being run by GitHub actions. This is a small docfix change to alert anyone trying to run the tests that they should use tox.

## Motivation and Context

It's important for contributors to know how to run the tests!

## Types of changes

## Checklist:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -62,7 +62,7 @@ To run and install locally you can:
 
 * Install with ``python setup.py install``
 
-* Run tests with ``python setup.py test``
+* Run tests with ``tox`` (you will need to ``pip install tox`` first)
 
 * Run Wayback with ``wayback`` (see docs for info on how to setup collections)
 


ikreymer:
I am able to get it to run via `python setup.py test`, though this command is deprecated, so should not recommend it.
Is plain `tox` what we want, though, as that runs in multiple python environments? For non-CI testing, probably just want to have it run in the current env?

ikreymer:
Looks like there's `https://pypi.org/project/tox-current-env` which allows you to run `tox --current-env` which runs tox in the current environment, if available, instead of all of them. I'll open a new PR that recommends using this approach

ikreymer:
Replacing with #754. @edsu if you have a chance, can you take a look at #754 to see if it makes sense / works for you?

Issue #727, opened 2022-06-17
HTML metadata.yaml is escaped


ArtHoff:
## Describe the bug
Hi 
I upgraded from pywb 2.4.0rc5 to the latest 2.6.7. 
Since the upgrade my metadata does not display properly in the interface.
I have entries in my metadata.yaml files that create links, and they used to be displayed as links and were clickable.
Now, after the upgrade the links are displayed as text.
When looking at the page source I can see that the code is encoded. Instead of <a href="blah"> it now shows &lt;a href=&#34;blah&#34;&gt;

## Steps to reproduce the bug
In the metadata.yaml file enter something like:
details: 'This site is being replaced by a new site: <a href="https://mysite.test.com/test">https://mysite.test.com/test</a>.'

## Expected behavior
I expect to see clikcable links like this:
![image](https://user-images.githubusercontent.com/24282952/174226270-5879c547-bc75-4fd7-9681-835876403ee6.png)

## Screenshots
![image](https://user-images.githubusercontent.com/24282952/174225283-c368e77b-acf4-47b4-aba0-a0a7724c68d8.png)

## Environment

- OS: RedHat Enterprise Linux Server release 7.9 (Maipo)
- Browser Firefox, Chrome, Edge
- Version not applicable

ArtHoff:
Hi,
Thank you for addressing this issue, however there is still an issue with the HTML rewrite.
On the home page the first URL is clickable, but the one in the comment is displayed as HTML.
I expect to see:
![Link](https://user-images.githubusercontent.com/24282952/206944291-a9101e58-2f81-41b6-8765-12deef245b40.jpg)

But instead it shows:
![No link](https://user-images.githubusercontent.com/24282952/206944204-a5bfb713-d4f0-4056-98cb-24f023965c7e.jpg)

Thanks for looking into this.

PR #747, opened 2022-07-29
Prevent jinja2 from escaping HTML markup in collection metadata

tw4l:
## Description

This PR adds jinja2 `safe` filters to metadata values and an `{% autoescape false %}` block to the collection title in the search template to prevent HTML markup in collection metadata from being escaped.

## Motivation and Context

Connected to https://github.com/webrecorder/pywb/issues/727

This fixes a regression in more recent versions of pywb which was reported by users.

### Before

(copied from linked issue)

image

### After

Screen Shot 2022-07-29 at 12 00 10 AM

## Types of changes

## Checklist:

### DIFF
--- a/pywb/templates/search.html
+++ b/pywb/templates/search.html
@@ -23,7 +23,9 @@ window.wb_prefix = "{{ wb_prefix }}";
             <div class="col-12">
                 <label for="search-url" class="lead" aria-label="Search For Col">
                     {% set coll_title = metadata.title if metadata and metadata.title else coll %}
-                    {% trans %}Search the {{ coll_title }} collection by url:{% endtrans %}
+                    {% autoescape false %}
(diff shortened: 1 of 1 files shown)


ikreymer:
@tw4l looks good, yeah I think this started happening with autoescape being defaulted to true..
Since this isn't tied to 2.7.0, mind rebasing this off main, and then we can just merge it in separately

tw4l:
@ikreymer no problem! this is now rebased and ready to merge

tw4l:
I hadn't noticed there was a new commit on `main` upstream - should be good now :)

ArtHoff:
Hi,
Thank you for addressing this issue, however there is still an issue with the HTML rewrite.
On the home page the first URL is clickable, but the one in the comment is displayed as HTML.

I expect to see:
image


But instead it shows:
image


Thanks for looking into this.

Issue #745, opened 2022-07-28
werkzeug 2.1.0 removal of useragents module causes error starting wayback


tw4l:
## Describe the bug

The 2.1.0 release of werkzeug [removed the `werkzeug.useragents` module](https://werkzeug.palletsprojects.com/en/2.1.x/changes/#version-2-1-0), which pywb uses in the `rewriterapp`. This causes the `wayback` command to throw a 500 Internal Server error.

Stacktrace:
```
pwyb_1  | Traceback (most recent call last):
pwyb_1  |   File "/usr/local/lib/python3.7/site-packages/pywb-2.7.0-py3.7.egg/pywb/apps/wayback.py", line 2, in <module>
pwyb_1  |     from pywb.apps.frontendapp import FrontEndApp
pwyb_1  |   File "/usr/local/lib/python3.7/site-packages/pywb-2.7.0-py3.7.egg/pywb/apps/frontendapp.py", line 26, in <module>
pwyb_1  |     from pywb.apps.rewriterapp import RewriterApp
pwyb_1  |   File "/usr/local/lib/python3.7/site-packages/pywb-2.7.0-py3.7.egg/pywb/apps/rewriterapp.py", line 12, in <module>
pwyb_1  |     from pywb.rewrite.default_rewriter import DefaultRewriter, RewriterWithJSProxy
pwyb_1  |   File "/usr/local/lib/python3.7/site-packages/pywb-2.7.0-py3.7.egg/pywb/rewrite/default_rewriter.py", line 23, in <module>
pwyb_1  |     from werkzeug.useragents import UserAgent
pwyb_1  | ModuleNotFoundError: No module named 'werkzeug.useragents'
```

## Steps to reproduce the bug

- Build the `2.7.0-beta` branch of pywb (likely other branches are affected as well), and run `wayback`, e.g. with docker-compose
- Go to `localhost:8080` in the browser
- See 500 Internal Server Error and inspect logs for traceback

## Environment

- OS: [e.g. iOS]: n/a
- Browser [e.g. chrome, safari]: n/a
- Version [e.g. 22]: Noticed in branch `2.7.0-beta`

## Additional context

The issue can be temporarily avoided by pinning werkzeug to `werkzeug==2.0.3` in the requirements file. For a more proper fix it looks like pywb should be modified to use the `werkzeug.user_agent` module instead.

ikreymer:
[...]

PR #749, opened 2022-07-29
Import UserAgent from werkzeug.user_agent rather than deprecated module

tw4l:
## Description

Werkzeug deprecated the `useragents` module in 2.0.x and removed it entirely in 2.1.x in favor of `werkzeug.user_agent`. This commit modifies the `rewriterapp` to import `UserAgent` from `user_agent` and sets requirements.txt to ensure versions of werkzeug prior to 2.0.x won't be used.

## Motivation and Context

Connected to #745 

The deprecation of `werkzeug.useragents` is causing a 500 Internal Server Error in the wayback machine.

## Types of changes

## Checklist:

### DIFF
--- a/pywb/rewrite/default_rewriter.py
+++ b/pywb/rewrite/default_rewriter.py
@@ -20,7 +20,7 @@ from pywb.rewrite.rewrite_js_workers import JSWorkerRewriter
 from pywb import DEFAULT_RULES_FILE
 
 import copy
-from werkzeug.useragents import UserAgent
+from werkzeug.user_agent import UserAgent
 
 
 # ============================================================================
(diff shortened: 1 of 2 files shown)


tw4l:
These changes seem to have caused another issue stemming from the pinned version of `markupsafe`:

```
#10 6.660 Installed /usr/local/lib/python3.7/site-packages/defusedxml-0.7.1-py3.7.egg
#10 6.703 error: MarkupSafe 2.0.1 is installed but MarkupSafe>=2.1.1 is required by {'werkzeug'}
```

@ikreymer if I remove `markupsafe<2.1.0` from pywb's requirements file, it builds/runs successfully again, but given that was added fairly recently in https://github.com/webrecorder/pywb/pull/697, maybe you can share a little context or advice if it's better to just pin `werkzeug==2.0.3` for now?

PR #751, opened 2022-08-16
Revisit headers load fix

ikreymer:
## Description
Revisit records may contain HTTP headers, or they may contain no headers. If they do contain headers, the replay should use the HTTP headers from the revisit record, and they payload from the original record referenced by the revisit.
Unfortunately, this was not working (possibly ever?) as expected!
It appears that pywb was always using the HTTP headers from the original record.
Generally, this is fine, except in the case where the revisit record is of a redirect, and the HTTP headers have changed between the revisit and the original (while the payload is the same, and generally ignored).

## Motivation and Context

This fixes an issue originally found in sul-dlss/was-pywb#64

Added test case which I believe replicates this issue.
cc: @edsu 

## Types of changes

## Checklist:

### DIFF
--- a/pywb/version.py
+++ b/pywb/version.py
@@ -1,4 +1,4 @@
-__version__ = '2.6.7'
+__version__ = '2.6.8'
 
 if __name__ == '__main__':
     print(__version__)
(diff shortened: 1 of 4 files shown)


edsu:
Thanks for investigating this!

PR #752, opened 2022-08-18
Enable translation for the remaining strings on the search results page

krakan:
A few strings on the search results pages were hard coded an not enclosed with the translation function which gave the page a half translated impression.

## Description

 * add the remaining strings to the `text` dict in `query.html`
     * `for matching` / `captures of`
     * `by`
     * `Mime Type`
     * `HTTP Status`
 * replace the occurrences of those strings with the corresponding dict element in `query.js`
 * tidy up the usage of bold in `query.js` for consistency
 * use the `tsToDate()` function also for the timestamps for search results without matchType
 * use the `document.documentElement.lang` parameter to decide how to localize the timestamps in `query.js`

The last item means that if the user has chosen a language for the page the timestamps will be formatted accordingly; otherwise the locale of the browser will control the formatting. It is totally feasible to use different language settings for the browser and the content.

## Motivation and Context

All other pages can be completely localized. This was as far as I can tell the remaining missing strings.

Before:
2022-08-18-093945_1049x219_scrot

After:
2022-08-18-091951_1048x216_scrot

(The National Library logo and the "Logga in" link are local customizations and not part of this change.)

## Types of changes

## Checklist:

The `python setup.py test` command fails early with a versioning conflict when running locally - with or without this change. I appears that the test subsystem does not care about version limitations in requirements.txt when installing python modules. As this is a non-functional change, there probably aren't any test for it anyway.

Trying to use localized ordinal suffixes for dates is a hard problem as there is no built in support for it in Javascript which I guess is why most timestamp-to-string implementations don't use ordinal suffixes.

### DIFF
--- a/pywb/static/query.js
+++ b/pywb/static/query.js
@@ -371,16 +371,13 @@ RenderCalendar.prototype.createContainers = function() {
         },
         { tag: 'textNode', value: ' ' },
         {
-          tag: 'b',
-          child: {
-            tag: 'textNode',
-            value: '',
-            ref: function(refToElem) {
-              renderCal.containers.versionsTextNode = refToElem;
-            }
+          tag: 'textNode',
+          value: '',
+          ref: function(refToElem) {
+            renderCal.containers.versionsTextNode = refToElem;
           }
         },
-        { tag: 'textNode', value: ' of ' + this.queryInfo.url }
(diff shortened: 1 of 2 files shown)


ikreymer:
Thank you for adding this!

PR #763, opened 2022-09-19
eval rewrite update + latest wombat (3.3.7)

ikreymer:
## Description

eval: switch to new eval rewriting which catches global scope rxrewriting: remove lookbehind check so that 'return eval(...)' can be rewritten tests: add additional eval tests

bump to 2.6.9

## Motivation and Context
Fixes rewriting of eval statements like `eval("import 'somefile.js'")`, using same system as in wabac.js and supported in wombat.js. Updates wombat.js to latest + eval rewriting logic to match wabac.js

## Types of changes

## Checklist:


pywb/rewrite/regex_rewriters.py
```diff
@@ -16,6 +16,19 @@ def remove_https(string, _):
...
         return lambda x, _: x.replace(match, replacer)
 
+    @staticmethod
+    def replace_prefix_from(prefix, match):
+        def do_replace(x, _):
+            start = x.find(match)
+            if start == 0:
+                return prefix
+            if start > 0:
+                return x[:start] + prefix
```
tw4l, line 26: This could be simplified to:

```python
if start >= 0:
    return x[:start] + prefix
```
since a `[:0]` slice of any string is just an empty string. Arguably it's a bit more readable as-is though, so maybe just leave it?

tw4l (approved):
Looks good to me! One nitpick but feel free to ignore it

pywb/rewrite/regex_rewriters.py
ikreymer: yeah, was thinking about readability, also ported it from:
https://github.com/webrecorder/wabac.js/blob/main/src/rewrite/jsrewriter.js#L38
to have some parity. so perhaps keep as is for now, unless changed in both places?

Issue #694, opened 2022-02-03
jsessionid suffix causes urls not to resolve in replay


ChrisDoyleMW:
## Describe the bug
Captured URLs that contain a jsession suffix do not replay

## Steps to reproduce the bug
1. Visit https://webarchive.nationalarchives.gov.uk/ukgwa/20130410190427/http://new.wales.gov.uk/funding/?lang=en
2. Click on Grants in the left hand menu.
3. This is a link to https://webarchive.nationalarchives.gov.uk/ukgwa/20130410190427mp_/http://new.wales.gov.uk/funding/fundgrantareas/;jsessionid=E3807C16B4795E305F208AD6315320AB?lang=en which does not replay
4. Edit the URL to remove **;jsessionid=E3807C16B4795E305F208AD6315320AB** and the page replays correctly

## Expected behavior
The links should work correctly. They work correctly if the ;jessionid string is removed. For example: https://webarchive.nationalarchives.gov.uk/ukgwa/20130410184640/http://new.wales.gov.uk/about/?lang=en. Note that the language string at the end: ?lang=en or ?lang=cy needs to remain as part of the urls.

## Screenshots
![Screenshot 2022-02-03 at 15 20 39](https://user-images.githubusercontent.com/38907291/152372029-fcbb0ade-d1a9-4a8f-95b4-d1a541f07320.png)
![Screenshot 2022-02-03 at 15 20 39](https://user-images.githubusercontent.com/38907291/152372131-615446f1-dd5e-4403-8d4c-d56f806815ca.png)
![Screenshot 2022-02-03 at 15 20 39](https://user-images.githubusercontent.com/38907291/152372296-50d5e7ab-0d05-438f-9f0f-e130b97ec52a.png)

## Environment

Replicated in Safari (Version 15.3 (16612.4.9.1.7, 16612)
Chrome (Version 97.0.4692.99 (Official Build) (x86_64)
Firefox (96.0.3 (64-bit)
OS - Windows 10 / MacOS / BigSur

## Additional context
This is the first issue I have raised - I hope all the above is sufficient - please ask me for anything further you need. Thanks, Chris

ChrisDoyleMW:
Here's what I've tried so far, I think I think the issue is down to a lack of understanding in how the rules.yaml works.

1. Local install of pywb 
2. Edited the config.yaml to load one of our collections
3. Checked the regex on regex101.com - https://regex101.com/r/lCzUCi/1 to make sure that it captures the correct part of the path
[...]

PR #780, opened 2022-11-29
feat: regex substitution on surt rules match

mijho:
## Description
substituion functionality already exists on a global level for matched rules but this causes issues when rule sets conflict in the desired outcome. This change enables setting regex substitution at the rule level to avoid these conflicts.

The only change to existing functionality is related to the `find_all` parameter moving to be an options in the `re_type` parameter. I've marked this down as a potentially breaking change as it may impact end users custom rule set, the default rules in the `rules.yaml` have been updated.

## Motivation and Context
We have experienced a number of issue that the rules.yaml functionality would resolve but due to substitutions being linked to the more global `url_normalize` functionality we couldn't make use of it fully.

An example issue this would resolve is #694. For this example a ruleset could be configured as below meaning that the `;jsession=foo` string is removed but the query parameters or any additional path data is retained.
```
    - url_prefix: uk,gov,wales,new)/
      fuzzy_lookup: 
        re_type: sub
        match: ';jsessionid=[0-9A-Z]{32}'
        replace: ''
```
In more complex circumstances match groups could be utilised to build the required url.

## Types of changes
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

### DIFF
--- a/pywb/rules.yaml
+++ b/pywb/rules.yaml
@@ -110,7 +110,7 @@ rules:
 
       fuzzy_lookup:
         match: '("(?:cursor|cursorindex)":["\d\w]+)'
-        find_all: true
+        re_type: findall
 
     - url_prefix: 'com,facebook)/ajax/pagelet/generic.php/profiletimeline'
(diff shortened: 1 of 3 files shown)


ikreymer:
Thanks, sorry for delay in merging this. Eventually, should probably combine the `url_normalize` functionality with this rewriting, but for now this should work.

PR #783, opened 2022-12-06
Make filter expressions translatable

krakan:
## Description

Add some more strings for translation by moving them from javascript files to the corresponding HTML templates
Also make pressing enter in the filtering expression field less surprising by adding a new filter instead of submitting the search without the filter being created. If no new filter is being created fall back to the previous behavior of submitting the form.

## Motivation and Context

All strings should be translatable to avoid annoying language mix ups. This PR fixes the remaining strings in `query.js` and `search.js`.
 
The query page where `Filtering by` and `Contains` et al. were previously not translatable:
2022-12-06-145945_1151x375_scrot
2022-12-06-150010_1154x378_scrot

## Types of changes

## Checklist:

### DIFF
--- a/pywb/static/query.js
+++ b/pywb/static/query.js
@@ -57,14 +57,6 @@ function RenderCalendar(init) {
   };
   // regex for extracting the filter constraints and filter mods to human explanation
   this.filterRE = /filter([^a-z]+)([a-z]+):(.+)/i;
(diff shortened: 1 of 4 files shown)


tw4l (approved):
Looks great, thank you @krakan! And for the screenshots as well :)

Issue #795, opened 2023-01-16
Show error when adding duplicate WARC


kuechensofa:
## Is your feature request related to a problem? Please describe.

When working with larger collections with many warc files it sometimes happens to me that I add a warc file with a filename that duplicates a already existing warc file in the collection. When I add it with wb-manager add [collection] [warc file] it just overwrites the existing file without asking or a warning.

## Describe the solution you'd like

Show an error when a warc file with a duplicate name should get added to a collection and abort the adding of the file.

PR #797, opened 2023-01-16
[#795] Show error when adding duplicate warc file

kuechensofa:
## Description

Show an error when adding a warc file that has the same filename as an existing file in the collection and do not overwrite the existing file.

## Motivation and Context

See #795 

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

### DIFF
--- a/pywb/manager/manager.py
+++ b/pywb/manager/manager.py
@@ -121,17 +121,23 @@ directory structure expected by pywb
                           format(self.archive_dir))
 
         full_paths = []
+        duplicate_warcs = []
         for filename in warcs:
             filename = os.path.abspath(filename)
 
+            # don't overwrite existing warcs with duplicate names
             if os.path.exists(os.path.join(self.archive_dir, os.path.basename(filename))):
-                raise IOError(f'Warc {filename} already exists')
+                duplicate_warcs.append(filename)
+                continue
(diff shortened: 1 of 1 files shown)


tw4l:
Thank you for the bug report and patch @kuechensofa! It would be great if instead of immediately raising an `IOError` when encounting a warc whose names already exists in the collection, we allowed the other warcs (those without naming conflicts) to be added and indexed, and print an error message at the end of the process listing any warcs that weren't able to be added.

Would you be able to make that change? Otherwise happy to do it myself.

kuechensofa:
> Thank you for the bug report and patch @kuechensofa! It would be great if instead of immediately raising an `IOError` when encounting a warc whose names already exists in the collection, we allowed the other warcs (those without naming conflicts) to be added and indexed, and throw the exception at the end of the process with the list of warcs that weren't able to be added.
> 
> Would you be able to make that change? Otherwise happy to do it myself.

Thanks for the reply! I'm happy to implement that change.

tw4l:
thanks @kuechensofa ! merged into `main` and will be included in the next point release :)

kuechensofa:
@tw4l Thanks, for merging! :)

Issue #804, opened 2023-01-31
Default Localisation misses on date format


oskarhek:
## Describe the bug
Default language is not used for formatting dates 

## Steps to reproduce the bug
for me the default language is set to NL
In the calendar when I select a date I still get the American notation with starts with a month.
after I select the nl language, it is transformed to day/month/year.

## Expected behavior
to see the correct format immediately (day/month/year)

![correct](https://user-images.githubusercontent.com/53567709/215757030-2b732df2-5247-45af-82a3-f190d3d43fc8.png)
![wrong](https://user-images.githubusercontent.com/53567709/215757041-e7a88700-e665-4c07-97c4-bf582fd91756.png)

krakan:
Yes, that is indeed the intended behavior. As I wrote this piece of code I'll have another go at getting it right but unfortunately "it works for me" ... :disappointed:

Update: Oh dear! I found the error - I've previously detected the issue and cheated in solving it meaning to fix it later and then promptly forgotten about it. The default language is indeed used - but only if it's Swedish. :blush:

oskarhek:
Can see if I can convince people to serve our pages in Swedish :)

tw4l:
Confirmed #805 resolves the issue - with default set to French and no language in request, calendar date is displaying DD/MM/YYYY as expected. Thanks both!

PR #805, opened 2023-01-31
[#804] Use default_locale when lang not set in the request

krakan:
Before a language is selected in a Pywb session it's expected that the `default_locale` from the config-file would be used. This works most of the time but not for eg. formatting dates on the query.html page where instead the `en` locale is used.

## Description

Let the FrontEndApp `__init__` method set a `default_locale` attribute as defined in the config file and use that as the default for incoming requests.

## Motivation and Context

See issue #804.

See issue #804.

## Types of changes

## Checklist:


tw4l:
Thanks for this @krakan ! :)

PR #812, opened 2023-02-10
Make empty URL:s work in ACL:s

krakan:
Setting a different default access for a specific user didn't previously work because

* the ACL must be sorted in reversed alphabetic order which means that an empty URL (ie. a default rule) must be last, and

* the search for matching ACL rules stops when the top level domain of the a rule is alphabetically lower than the one searched for.

Ie. given `default_access: allow` and the ACL:

```
com,example)/ - {"access": "block"}
 - {"access": "block", "user": "unknown"}
```

and searching for `http://iana.org`, the last line would not be found so that the `unknown` user would still be `allowed`.

## Description

Don't stop searching for rules based on top level domain

## Motivation and Context

Having a way to specify a default access rule for a specific user would be useful. In our case we would give external users access to the index but not to the data. This was previously possible only if the rule with an empty URL was the only one.

## Types of changes

## Checklist:


ikreymer:
This is a neat solution for default rules by user, the system wasn't really designed for that as is.
My main concern is continuing the search just to find the blank entry, if there are a lot of exclusion rules, it would slow down the search.

It may be clearer to disallow empty URLs, but instead support a list of default rules by user:

```
default_access:
   # default user
   - user: ""
      access: block
      
   - user: member
      access: allow
```

and then just check from this list instead?

krakan:
Agreed, that is indeed clearer - I'll look into changing the PR.

krakan:
I opted for a dict layout of the default_access entry instead of the proposed list:
```
default_access:
   default: block
   member: allow
```

krakan:
There! I had forgotten to add a file (again :facepalm:) - and I also rebased on the `origin/dev/skip-youtube-dl-test-in-ci` branch

Issue #715, opened 2022-05-10
ACL access allow_ignore_embargo doesn't work with a user


krakan:
## Describe the bug

When specifying an embargo for a collection and then putting allow_ignore_embargo for a specific user in the aclj-file, the embargoed url:s don't show up in the search results. If one circumvents the search results and goes directly to the page view, the page actually shows up.

## Steps to reproduce the bug

config.yaml:

```
collections:
  exempel:
    archive_paths: /usr/local/sample_archive/warcs/
    index: /usr/local/sample_archive/cdx
    embargo:
      newer:
        years: 10
    acl_paths: allows.aclj
```
allows.aclj:

```
com,example)/ - {"access": "allow_ignore_embargo", "user": "admin"}
```
## Actual behavior
```
curl -iHX-Pywb-ACL-User:admin 'http://localhost:8090/exempel/cdx?url=example.com&output=json'
HTTP/1.1 200 OK
Content-Type: text/x-ndjson


```
## Expected behavior

```
curl -iHX-Pywb-ACL-User:admin 'http://localhost:8090/exempel/cdx?url=example.com&output=json'
HTTP/1.1 200 OK
Content-Type: text/x-ndjson

{"urlkey": "com,example)/", "timestamp": "20130729195151", "url": "http://test@example.com/", "mime": "warc/revisit", "status": "-", "digest": "B2LTWWPUOYAH7UIPQ7ZUPQ4VMBSVC36A", "redirect": "-", "robotflags": "-", "length": "591", "offset": "355", "filename": "example-url-agnostic-revisit.warc.gz", "source": "exempel:url-agnost-example.cdx", "source-coll": "exempel", "access": "block"}
{"urlkey": "com,example)/", "timestamp": "20140127171200", "url": "http://example.com", "mime": "text/html", "status": "200", "digest": "B2LTWWPUOYAH7UIPQ7ZUPQ4VMBSVC36A", "redirect": "-", "robotflags": "-", "length": "1046", "offset": "334", "filename": "dupes.warc.gz", "source": "exempel:dupes.cdx", "source-coll": "exempel", "access": "block"}
[...]

PR #813, opened 2023-02-13
[#715] Forward custom headers for cdx queries

krakan:
## Description
Find custom http headers in the incoming request and forward them to the cdx-request. A custom header in this context is one prefixed with `HTML_X_`.

## Motivation and Context

The difference between the `block` and `exclude` actions in the ACL is that the former should allow access to the index but not the data whereas the latter should block both index and data access.

See issue [#715] which is about `allow_ignore_embargo` not allowing access to the index when specified with a specific `user`. On investigation the issue also affected other use cases. The change set includes new tests for theses cases.

## Types of changes

## Checklist:

### DIFF
--- a/pywb/apps/frontendapp.py
+++ b/pywb/apps/frontendapp.py
@@ -434,7 +434,11 @@ class FrontEndApp(object):
             cdx_url += 'limit=' + str(self.query_limit)
 
         try:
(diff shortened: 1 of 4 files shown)


tw4l (approved):
Looks good, thank you!

PR #820, opened 2023-02-28
Sort index when adding wacz archives

kuechensofa:
## Description

Ensure that collection index is sorted when adding wacz archives to the collection.

## Motivation and Context

CDXJ indices must be sorted for the binary search algorithm to work. When adding wacz archives and merging there indices with the collection index, the wacz index was just appended to the end of the collection index and the collection index wasn't sorted anymore.

## Types of changes

## Checklist:


tw4l (approved):
May want to tweak before 2.8, but merging so we don't have a broken feature in main. Thanks for the contribution!

PR #841, opened 2023-04-24
Documentation: Expand instructions for Certificate Authority setup

ntevenhere:
## Motivation and Context

I found the documentation for using pywb Certificate Authority difficult to use for adding the CA to Chromium, as I had never done something like that before. I attempted to make the 4th step as clear as possible, "4. You may need to agree to "Trust this CA" to identify websites." (Agree where?) Also, the ordering of the paragraphs was somewhat disjointed.

### DIFF
--- a/docs/manual/configuring.rst
+++ b/docs/manual/configuring.rst
@@ -549,17 +549,18 @@ using the special download paths. Recommended set up for using the proxy is as f
 
 2. Configure the browser proxy settings host port, for example ``localhost`` and ``8080`` (if running locally)
 
-3. Download the CA:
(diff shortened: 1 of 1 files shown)


notevenaperson:
These are some links that could would be useful in the instructions, but I omitted them to keep the documentation brief, and also free from hand-waving solutions to untested resources.

- https://support.mozilla.org/en-US/kb/setting-certificate-authorities-firefox
- https://stackoverflow.com/questions/7580508/getting-chrome-to-accept-self-signed-localhost-certificate/15076602

PR #868, opened 2023-10-26
Sets "Pywb Error" string as translatable in error template

### LINKED ISSUE #867: String not set as translatable in template
## Describe the bug
In the error.html template, on line 6, "Pywb Error" inside the `<h2>` tag isn't set to be translatable. [Link to the line](https://github.com/webrecorder/pywb/blob/main/pywb/templates/error.html#L6).

The translation string for it already exists because it's set to be translatable just a few lines above (when setting the page title).

## Steps to reproduce the bug
Clone new repository, set up translations as per the documentation, change the "Pywb Error" string in translation file, open the Pywb site, change translation language to the one you translated, open an error page by using a collection link that doesn't exist (has to include the translation language code in the URL), the title "Pywb Error" remains the same (not translated).

## Expected behavior
The heading "Pywb Error" is translated when the error page opens. Just like the page title.

## Screenshots
image
Notice how the other text is translated (including the page title), but the "Pywb Error" heading isn't. The URL in the screenshot is: http://localhost:8080/hr/dsfgd/

## Environment
- Debian 10 WSL2 server, Windows 11 client
- Pywb version 2.7.4 + i18n
- Python version 3.9.2
- Chrome version 118.0.5993.89 (Official Build) (64-bit)

Quirinus:
## Description
Sets the "Pywb Error" string in the `<h2>` heading as translatable in the error.html template.

## Motivation and Context
- The string "Pywb Error" isn't translatable [in this line](https://github.com/webrecorder/pywb/blob/main/pywb/templates/error.html#L6).
- Fixes #867 

image
The "Pywb Error" isn't translated, but this pull request fixes that.

## Types of changes

## Checklist:

I have no idea how to run the tests, but hopefully just making a string translatable shouldn't break anything.

### DIFF
--- a/pywb/templates/error.html
+++ b/pywb/templates/error.html
@@ -3,7 +3,7 @@
 {% block body %}
 <div class="container text-danger error">
     <div class="row justify-content-center">
-        <h2 class="display-2">Pywb Error</h2>
+        <h2 class="display-2">{{ _('Pywb Error') }}</h2>
     </div>
     <div class="row">
         <div class="col-12 text-center">


tw4l (approved):
Thank you for the fix!

tw4l:
Confirmed that the test failure seen in CI is happening on main and not related to this change (thought so but wanted to make sure!), so I'm going to go ahead and merge this in and resolve that separately.

PR #871, opened 2023-11-02
Fixes switch_locale not adding locale if missing from URL

### LINKED ISSUE #870: switch_locale not adding locale if missing from URL
## Describe the bug
If switch_locale(locale) is used and the URI doesn't have the language code in it, it will produce URLs without it. That means if the URL is e.g. https://localhost:8080/test/ and switch_locale("hr") or switch_locale("en") is used, it will give https://localhost:8080/test/ instead of https://localhost:8080/hr/test/. It works fine if the language code is in the URI - https://localhost:8080/hr/test/ and switch_locale("en") produces https://localhost:8080/en/test/.

## Steps to reproduce the bug
Install pywb and i18n package, add two languages for translations, open a pywb page without the language code in the URI. I have "en" and "hr", with "hr" set as default in the config.yaml. The language switch links produced using {{ switch_locale(locale) }} in the template in the top right will point to the same URI (for "hr" and "en" both), without the language codes - so it won't switch the language.

## Expected behavior
If switch_locale(locale) is used with a URI that doesn't have the language code, it should produce URLs that contain the language code. That means the language switch links in the top right of pages should contain the correct URLs and switch languages accordingly.

## Screenshots
image

Notice the current URL for a collection named "test" (not containing any language code in the URL). I hovered the mouse over the link to change the locale to "hr" (top right), and the URL that is in that link is shown on bottom left - it doesn't contain "hr". Same happens when I hover over "en".

## Environment

- OS: Windows 11 and Ubuntu 10
- Chrome 118.0.5993.89 (Official Build) (64-bit)
- Pywb 2.7.4

## Additional context
None.

Quirinus:
If the language code was missing in the URI, switch_locale(locale) didn't add it (it worked fine if it was present). That means that it produced the same URL for all locales, each missing the language code in the URL.

## Description
I added an additional check that checks if the language is set in the URI, and it is changed only in that case. Otherwise if it was missing, it was producing a URL that doesn't contain it - but now it does add the launguage code to the URL if it was missing.

## Motivation and Context
Fixes #870

image

Notice the current URL for a collection named "test" (not containing any language code in the URL). I hovered the mouse over the link to change the locale to "hr" (top right), and the URL that is in that link is shown on bottom left - it doesn't contain "hr". Same happens when I hover over "en".

## Types of changes

## Checklist:

I don't know how to run the tests.

### DIFF
--- a/pywb/rewrite/templateview.py
+++ b/pywb/rewrite/templateview.py
@@ -178,7 +178,7 @@ class JinjaEnv(object):
 
             request_uri = environ.get('REQUEST_URI', environ.get('PATH_INFO'))
 
-            if curr_loc:
+            if curr_loc and request_uri.startswith('/' + curr_loc + '/'):
                 return request_uri.replace(curr_loc, locale, 1)
 
             app_prefix = environ.get('pywb.app_prefix', '')


Quirinus:
I tested this using the default web server ("wayback" command). Hopefully it works well with uwsgi too.

I will check when I have time.

tw4l (approved):
Tested and working well, thank you!

PR #873, opened 2023-11-07
Fixes environ paths when default_locale set

### LINKED ISSUE #872: No search results (by domain) if default_locale set
## Describe the bug
If you set default_locale in config.yaml, and visit search results (by domain) with no language code in the URL path, it will not return any results.

## Steps to reproduce the bug
Install pywb, the i18n package, add a collection. Create a translation, add it to the locale in config.yaml and set default_locale to that language. Run pywb with "wayback". Make a search by domain, and then remove the language code part from the URL, if there is any, and refresh. There will be no results.

## Expected behavior
Search results are shown.

## Environment
- pywb version 2.7.4
- using the default web server "wayback"
- Debian 11

Quirinus:
## Description
If the default_locale was set in config.yaml and the URL path didn't contain a language code, it was behaving as if there was a language code in the URL. In that case, it was moving part of the PATH_INFO to SCRIPT_NAME, but as there wasn't any language code in the URL path, it moved something else. This fixes that.

## Motivation and Context
I noticed this bug when the search results (by domain) didn't show any results, in the condition described above.
This fixes #872

## Types of changes

I marked it as bug fix, instead of a breaking change, but I'm not 100% sure.

## Checklist:

I don't know how to run tests.

### DIFF
--- a/pywb/apps/frontendapp.py
+++ b/pywb/apps/frontendapp.py
@@ -667,10 +667,14 @@ class FrontEndApp(object):
             # store original script_name (original prefix) before modifications are made
             environ['ORIG_SCRIPT_NAME'] = environ.get('SCRIPT_NAME')
 
-            lang = args.pop('lang', self.default_locale)
+            lang = args.pop('lang', '')
             if lang:
                 pop_path_info(environ)
+
+            if lang:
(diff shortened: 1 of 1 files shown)


tw4l (approved):
Thank you!

PR #882, opened 2024-02-13
Allow ACLJs to use *, SURT wildcard to match all URLs

### LINKED ISSUE #881: Allow for catch-all wildcard *, in ACLs
## Is your feature request related to a problem? Please describe.

Currently, if a user wants to set `default_access: block` but allow particular users to have access to any URL, there is no way to do this other than enumerating all of the possible TLDs in their collection in an ACLJ file, e.g.:

```
com, - {"access": "allow", "user": "staff"}
org, - {"access": "allow", "user": "staff"}
...
```

## Describe the solution you'd like

I'd like a user to be able to specify an ACL rule matching any possible URL, with the following syntax:

```
*, - {"access": "allow", "user": "staff"}
```

## Describe alternatives you've considered

- Requiring users to list all TLDs (cumbersome)
- Simply using `*` rather than `*,` as the key to match any URL (doesn't match existing syntax)

## Additional context

This originally came in as a user request on IIPC Slack.

tw4l:
Fixes #881

## Description

This PR adds the option to use `*,` as a SURT in ACLJ files to match any URL. This builds off of the existing ability to do this per-TLD, e.g. `com,`.

The PR also adds a passing test for the use case described in the issue (overriding `default_access: block` to all URLs for a specific user), and documents the new option.

## Motivation and Context

This allows administrators to give access to all resources to specific (e.g. staff) users.

## Types of changes

## Checklist:

- [X] I have added or updated tests to cover my changes.
- [x] All new and existing tests passed

### DIFF
--- a/docs/manual/access-control.rst
+++ b/docs/manual/access-control.rst
@@ -105,6 +105,10 @@ Given these rules, a user would:
 * but would receive an 'access blocked' error message when viewing ``http://httpbin.org/`` (block)
 * would receive a 404 not found error when viewing ``http://httpbin.org/anything`` (exclude)
 
+To match any possible URL in an .aclj file, set ``*,`` as the leading SURT, for example::
+
+  *, - {"access": "allow"}
+
 
 Access Types: allow, block, exclude, allow_ignore_embargo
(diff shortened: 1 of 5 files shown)


ikreymer (approved):
Nice! Looks good, only thought is if its worth mentioning that the `*,` rule should generally be last, which works out well since it sorts correctly and would be before any valid domain.

PR #888, opened 2024-03-21
rewrite: stop prepending semicolon to `this.` special property access…

### LINKED ISSUE #850: jquery rewriting causes error
## Description

Since an update of pywb I have an object with a jquery problem. It seems to be a rewritting issue.
The issue occurs since pywb 2.6.9. Everything workes fine in pywb 2.6.8., ReplayWeb.Page and Conifer. 

live web: https://dejame-c0mer.tumblr.com/
Pywb Replay (2.7.3): https://webarchives.rhizome.org/Q15390/20230111102538/https://dejame-c0mer.tumblr.com/
Conifer: https://conifer.rhizome.org/mona/dejame-c0mer/list/pages-detected/b1/20230111102655/https://dejame-c0mer.tumblr.com/

## Expected behavior

As you can see in the live web or conifer replay, draggable divs appear on the web page. 

## What actually happened

1. Go to https://webarchives.rhizome.org/Q15390/20230111102538/https://dejame-c0mer.tumblr.com/

Draggable divs are missing in pywb replay.

## Error Messages

Error message in dev console (pywb replay):
    (1) Uncaught SyntaxError: expected expression, got ';’
    (2) Uncaught TypeError: $(...).draggable is not a function

The first error message refers to the resource jquery-ui.js and this particular part:

```
if ( element ) {
    element = element.jquery || element.nodeType ?
        $( element ) :
        ;_____WB$wombat$check$this$function_____(this).document.find( element ).eq( 0 );
}
```

## Browser
**Desktop:**
- OS: Ubuntu 22.04
- Browser: Firefox
- Version 114.0.1

ato:
Fixes #850

## Description

This changes removes the regex rewrite rule that prepends a semicolon when `this.` special property access occurs immediately after a newline.

For example:

```js
x = 1 + 2
this.location = "foo"
```

will now be rewritten as:

```js
x = 1 + 2
_____WB$wombat$check$this$function_____(this).location = "foo"
```

instead of:

```js
x = 1 + 2
;_____WB$wombat$check$this$function_____(this).location = "foo"
```

## Motivation and Context

The prepended semicolon breaks code (such as jQuery in #850) that looks like:

    foo = foo ? foo :
                this.location;

I think the reason we started inserting the semicolon was because in situations like:

    x = 1 + 2
    this.location = "foo"

we used to rewrite to:

    x = 1 + 2
    (this && this._WB_wombat_obj_proxy || this).location = "foo"

which the browser would interpret as a bogus function call like `2(this && ... )`. But nowadays prepending the semicolon should be unnecessary as we currently rewrite to:

    x = 2 + 3
    _____WB$wombat$check$this$function_____(this).location = "foo"

which will trigger JavaScript's automatic semicolon insertion rules like the original code does.

## Types of changes

## Checklist:

### DIFF
--- a/pywb/rewrite/regex_rewriters.py
+++ b/pywb/rewrite/regex_rewriters.py
@@ -124,9 +124,7 @@ if (!self.__WB_pmw) {{ self.__WB_pmw = function(obj) {{ this.__WB_source = obj; 
             (r'(?<![$.])\s*\blocation\b\s*[=]\s*(?![=])', self.add_suffix(check_loc), 0),
             # rewriting 'return this'
             (r'\breturn\s+this\b\s*(?![.$])', self.replace_str(this_rw), 0),
-            # rewriting 'this.' special properties access on new line, with ; prepended
-            (r'\n\s*this\b(?=(?:\.(?:{0})\b))'.format(prop_str), self.replace_str(';' + this_rw), 0),
-            # rewriting 'this.' special properties access, not on new line (no ;)
+            # rewriting 'this.' special properties access
             (r'(?<![$.])\s*this\b(?=(?:\.(?:{0})\b))'.format(prop_str), self.replace_str(this_rw), 0),
             # rewrite '= this' or ', this'
             (r'(?<=[=,])\s*this\b\s*(?![:.$])', self.replace_str(this_rw), 0),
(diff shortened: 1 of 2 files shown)


tw4l:
Hi @ato , now that we've worked out the dependency upgrades and related test issues in main, would you mind rebasing this on latest main? Thanks!

ato:
Rebased. :-)

ikreymer:
Thanks for this @ato (and remembering the history of the adding the semicolon).

In wabac.js, we made this a more involved check to see if it the previous character is actually a newline, and a semicolon only then:
https://github.com/webrecorder/wabac.js/blob/main/src/rewrite/jsrewriter.js#L76
from matching:
https://github.com/webrecorder/wabac.js/blob/main/src/rewrite/jsrewriter.js#L115
so we don't have that issue there, but maybe even that is unnecessary?

ikreymer (approved):

PR #893, opened 2024-04-03
Use native JSON values in query string for POST canonicalization

### LINKED ISSUE #859: Canonicalize non-GET URLs with native JSON values
The way that pywb rewrites URLs for POST (and other non-GET) request canonicalization ends up writing Pythonic values into the URL such as `True`, `False`, and `None`, whereas we ideally want to have it use valid JSON values.

This is part of making POST canonicalization consistent across Webrecorder tools. Related to https://github.com/webrecorder/specs/issues/141

tw4l:
Fixes #859 

## Description

Modifies the output of POST canonicalization query strings in pywb to use native JSON values for booleans, numbers, and null, rather than a string representation of their Python values.

This commit also adds a more complicated JSON test case that is also in warcio.js to ensure parity.

We now handle numbers like JavaScript's Number.prototype.toString() by dropping decimal from floats if they represent whole number, to ensure consistency between pywb and warcio.js.

## Motivation and context

This is part of a cross-repo effort to standardize how POST canonicalization works in Webrecorder tools, and document this in a Webrecorder specfiication.

## Testing notes

Seems to be working well with fuzzy matching on replay, but could use some additional testing.

## Types of changes

## Checklist:

- [] My change requires a change to the documentation

### DIFF
--- a/pywb/warcserver/inputrequest.py
+++ b/pywb/warcserver/inputrequest.py
@@ -11,6 +11,7 @@ from io import BytesIO
 import base64
 import cgi
 import json
+import math
 import sys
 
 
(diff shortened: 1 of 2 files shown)


ikreymer (approved):
Looks good! Maybe add one more test that involves the float conversion, otherwise should be good to go!

PR #902, opened 2024-04-24
Handle WARC filename conflicts with wb-manager add

tw4l:
## Description

Instead of not copying a WARC to the archive directory in case of a filename conflict, this PR instead modifies it to add the WARC with `{original_name}-{dupe index}{extension}` - i.e. `example.warc.gz` would be renamed to `example-1.warc.gz`, then `example-2.warc.gz`, and so on, until a filename that doesn't exist in the archive directory is found.

## Motivation and Context

Should make for a more user-friendly experience of adding archives via `wb-manager add`.

## Types of changes

## Checklist:


tw4l:
Approved offline at WAC!

Quirinus:
Thanks!

Issue #913, opened 2024-08-18
`datetime.datetime.utcnow()` is deprecated since Python 3.12


Lisias:
I'm running `pywb` on Python 3.12 nowadays, and I had noticed the following message on the logs now and then:

```
Deprecated since version 3.12: Use datetime.now() with UTC instead.
```

Researching about, I learnt that `datetime.datetime.utcnow()` is being deprecated - if I understood correctly, naive `datetime`s are going to be avoided.

Since is unavoidable that `utcnow()` will be removed from the Python's library, I consider prudent to prevent users from being caught by surprise in the foreseeable future.

Lisias:
Done. All my processes were rebooted using WARCIO 1.7.5 and PYWB with the patches from this [pull request](https://github.com/webrecorder/pywb/pull/914) applied.

PR #914, opened 2024-08-18
Preventing the Python 3.12's "Deprecated since version 3.12: Use datetime.now() with UTC instead."

Lisias:
## Description

See https://github.com/webrecorder/pywb/issues/913

## Types of changes

## Checklist:

### DIFF
--- a/pywb/warcserver/access_checker.py
+++ b/pywb/warcserver/access_checker.py
@@ -7,7 +7,7 @@ from pywb.utils.binsearch import search
 from pywb.utils.merge import merge
 
 from warcio.timeutils import timestamp_to_datetime
-from datetime import datetime, timedelta
+from datetime import datetime, timezone
 from dateutil.relativedelta import relativedelta
 import os
 
@@ -164,13 +164,13 @@ class AccessChecker(object):
         # embargo if newser than
         newer = self.embargo.get('newer')
         if newer:
-            actual = datetime.utcnow() - newer
+            actual = datetime.now(timezone.utc) - newer
             return access if actual < dt else None
 
         # embargo if older than
         older = self.embargo.get('older')
         if older:
-            actual = datetime.utcnow() - older
+            actual = datetime.now(timezone.utc) - older
             return access if actual > dt else None
 
     def create_access_aggregator(self, source_files):
--- a/pywb/recorder/multifilewarcwriter.py
+++ b/pywb/recorder/multifilewarcwriter.py
@@ -245,7 +245,7 @@ class MultiFileWARCWriter(BaseWARCWriter):
         if not self.max_idle_time:
             return
 
-        now = datetime.datetime.now()
+        now = datetime.datetime.now(datetime.timezone.utc)
 
         for dir_key, out, filename in self.iter_open_files():
             try:
@@ -254,7 +254,7 @@ class MultiFileWARCWriter(BaseWARCWriter):
                 self.close_key(dir_key)
                 return
 
-            mtime = datetime.datetime.fromtimestamp(mtime)
+            mtime = datetime.datetime.fromtimestamp(mtime,tz=datetime.timezone.utc)
 
             if (now - mtime) > self.max_idle_time:
                 print('Closing idle ' + filename)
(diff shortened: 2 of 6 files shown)


Lisias:
I made a mistake, forgot that WARCIO is also using naive datetimes. I'm working on a short term solution, please wait.

asfia10-1996 (approved):

Lisias:
I finally got my stuff together and remembered this one.

I just realized that WARCIO was updated:

https://github.com/webrecorder/warcio/pull/185

(yeah, I'm a bit slow... 😬 )

I'm revising this code and comparing it with whatever I have something in (my custom) production at this moment - that code is working for months already, time more than enough to had bitten my back if I had made some serious mistake.

Lisias:
That's the history: With WARCIO 1.7.5 giving us the option for getting non naive `datetime`s from it, after this pull request is applied we would have internally code using both `datetime` types: the deprecating naive ones, and the new canon tz aware ones.

This would be a problem on code touched by the previous commit, and so code reviewing the changes I found that `pywb.warcserver.access_checker` and `pywb.warcserver.resource.responseloader` would end up handling both times concurrently - what could lead to undesired surprises on the long run.

So I changed these, adding `True` to the calls to WARCIO's `timestamp_to_datetime` at:

* https://github.com/webrecorder/pywb/blob/7b0f8b58607fb0ed338f0cfddeb80c629582d8f6/pywb/rewrite/templateview.py#L247
* https://github.com/webrecorder/pywb/blob/7b0f8b58607fb0ed338f0cfddeb80c629582d8f6/pywb/warcserver/access_checker.py#L118
* https://github.com/webrecorder/pywb/blob/7b0f8b58607fb0ed338f0cfddeb80c629582d8f6/pywb/warcserver/access_checker.py#L122
* https://github.com/webrecorder/pywb/blob/7b0f8b58607fb0ed338f0cfddeb80c629582d8f6/pywb/warcserver/resource/responseloader.py#L305
* https://github.com/webrecorder/pywb/blob/7b0f8b58607fb0ed338f0cfddeb80c629582d8f6/pywb/warcserver/resource/responseloader.py#L585

Doing a second code review, this time worrying about the consequences of the changes, I realized that these changes **are not needed** because the `datetime` generated are fed back into WARCIO and then forgot about - and so there's no possible clash between naive and tz aware `datetime`s - and this last commit of mine is useless.

However, I had forgot that updating my branch our reflect automatically on this pull request, my bad. So I decided to rebase by branch back one commit, and ask first if you guys would prefer to have the code above changed due some internal compliance, or if whatever we have now is good to go.

I'm marking this pull request as Ready for Review anyway - as I found no possible problem with WARCIO with or without getting naive `datetimes` from them.

pywb/warcserver/access_checker.py
```diff
@@ -164,13 +164,13 @@ def check_embargo(self, url, ts):
         # embargo if newser than
         newer = self.embargo.get('newer')
         if newer:
-            actual = datetime.utcnow() - newer
+            actual = datetime.now(timezone.utc) - newer
```
tw4l, line 167:
```suggestion
            actual = datetime.now(timezone.utc).replace(tzinfo=None) - newer
```

I think this is one of the few places where we do need to explicitly create a naive datetime or else an exception will be thrown due to comparison of naive and aware datetimes.

pywb/warcserver/access_checker.py
```diff
@@ -164,13 +164,13 @@ def check_embargo(self, url, ts):
...
             return access if actual < dt else None
 
         # embargo if older than
         older = self.embargo.get('older')
         if older:
-            actual = datetime.utcnow() - older
+            actual = datetime.now(timezone.utc) - older
```
tw4l, line 173:
```suggestion
            actual = datetime.now(timezone.utc).replace(tzinfo=None) - newer
```

Same as above.

tw4l (commented):
Thank you for this PR, and apologies for our taking a while to review it! I think there are two cases where we explicitly need to create naive datetimes still to avoid comparison issues.

pywb/warcserver/access_checker.py
wumpus: `newer` is a relativedelta, not a datetime.

tw4l: You're absolutely right, my apologies!

tw4l (approved):

pywb/warcserver/access_checker.py
tw4l: Ah that's right, the issue wasn't with `newer` but with the datetime created from a timestamp that `actual` is compared against. Fix in https://github.com/webrecorder/pywb/pull/946

Issue #865, opened 2023-09-22
Pywb failing to handle self-redirects from OutbackCDX


obrienben:
## Describe the bug
Pywb is throwing a LiveResourceException when receiving a self-redirect (3xx) from OutbackCDX. This results in Pywb displaying a blank page with the text "Not found".

## Steps to reproduce the bug
Example warc file attached in this Slack thread
https://iipc.slack.com/archives/C2NR32PNF/p1691445882952669
Try accessing "http://2020.org.nz/" using the redirect record, should display "Not found" message.

## Expected behavior
Pywb to process the self-redirect record from OutbackCDX, and load the record that the self-redirect points to.

## Screenshots
Pywb logs
![image](https://github.com/webrecorder/pywb/assets/5760612/a7810db3-c577-4f71-8538-bde8f0d2ff54)
OutbackCDX logs
![image](https://github.com/webrecorder/pywb/assets/5760612/4cb5b830-8a3a-4333-8973-8110bb61ab68)

## Environment
- Not browser specific
- Occurs in Pywb v2.7.0 onwards
- Does not occur before v2.7.0
- OutbackCDX v0.11.0

wumpus:
Thank you for the excellent bug report, with the pywb version dependence.

I can't access the slack warc file because my org isn't a member. I only have guest access.

lasztoth:
We experience the same issue with OutbackCDX v. 0.11.1 and PyWb v. 2.7.4. Redirects result in "Not found".

obrienben:
@wumpus unfortunately the warc was too big to attach here. Happy to share it with you another way if you'd like it

wumpus:
I see Ilya has been assigned by Tessa and I know he does have access to the IIPC Slack. So it's in good hands.

HeliosLHC:
Echoing same error as well with PyWb 2.83 and OutbackCDX 1.0.0

ato:
OutbackCDX has a partial workaround for this. If you run it with the --omit-self-redirects command-line option (or pass omitSelfRedirects=true in the query string) it will try to use the CDX redirect field to detect self redirects and hide them.

Unfortunately pywb's cdx-indexer and webrecorder/cdxj-indexer don't populate the redirect field though so if you used them to build your indexes this workaround won't help you. Without the redirect field populated there's no way for OutbackCDX to detect self redirects.

[...]

PR #917, opened 2024-08-22
A fix for GitHub issue #865

lasztoth:
## Description
This PR fixes the issue #865. Specifically, it appears that the bug was introduced (perhaps by mistake) by commenting out lines 288-289 from `responseloader.py`. It appears that uncommenting these lines, i.e., returning from the method if there is already a WARC filename and offset for the record, then the self-redirects work correctly. These lines were commented between versions 2.6.9 and 2.7.0b, which corresponds to the issue description. After this bug fix, self-redirects work correctly once again (on a local test system).

## Motivation and Context
Solves #865.

## Types of changes

## Checklist:

### DIFF
--- a/pywb/warcserver/resource/responseloader.py
+++ b/pywb/warcserver/resource/responseloader.py
@@ -285,8 +285,8 @@ class LiveWebLoader(BaseLoader):
             self.socks_proxy = None
 
     def load_resource(self, cdx, params):
-        #if cdx.get('filename') and cdx.get('offset') is not None:
-        #    return None
+        if cdx.get('filename') and cdx.get('offset') is not None:
+            return None
 
         load_url = cdx.get('load_url')
         if not load_url:


tw4l:
I have a vague memory now of this being commented out in relation to the development of the Vue banner UI, but for the life of me can't remember why. Will try to dig back through my notes from that time!

tw4l:
Ah, I dug back through our internal Discord conversations and remember now! This was our hacky temporary solution at the time to get pywb to populate the calendar from a remote CDX server of form `cdx+https://example.com/webarchive/cdx` in pywb 2.6.8, as without these lines commented out we'd receive `No such file or directory` errors because the actual archive files weren't available locally. It likely got committed as part of the Vue banner work by accident, as the PR we ended up merging at the time for what was very large.

I believe it should be safe to uncomment these in 2.7+ but we'll want to do some testing of course.

lasztoth:
Thanks a lot @tw4l for taking a look at this!

obrienben:
@tw4l are you looking into a fix that addresses the calendar issue you mentioned and [865](https://github.com/webrecorder/pywb/issues/865)?

tw4l:
Hi @lasztoth and @obrienben, I'm having some health issues that have limited my capacity at the moment. I'll be out of the office next week but will try to prioritize this when I return.

obrienben:
@tw4l hope you're doing better. Is there any update on this PR?

tw4l:
Hi, thanks @obrienben! This is in scope for the IIPC-funded pywb work, which is due to be delivered end of March, so we will be working on this in the coming weeks. Thanks for your patience!

ikreymer:
Closing in favor of #959 which includes this fix but as a conditional to avoid breaking for remote CDX/live web (see PR for more details).

PR #933, opened 2025-04-23
Fix tests, support py3.9, 3.10, 3.11

ikreymer:
## Types of changes

## Checklist:


pywb/utils/loaders.py
```diff
@@ -352,11 +353,13 @@ def load(self, url, offset, length):
             else:
                 headers['Cookie'] = self.cookie_maker.make()
 
-        if not self.session:
-            self.session = requests.Session()
+        #if not self.session:
```
tw4l, line 356:
```suggestion
```

pywb/utils/loaders.py
```diff
@@ -336,6 +336,7 @@ def __init__(self, **kwargs):
         if not self.cookie_maker:
             self.cookie_maker = kwargs.get('cookie')
         self.session = None
```
tw4l, line 338:
```suggestion
```

tw4l: Maybe we just remove this altogether if we're no longer using it

ikreymer: oh, i didn't mean to make that change for now, it didn't help with previous issues. will just undo, good catch!

Issue #947, opened 2025-07-23
Saving zstd-encoded content fails.


0x1b5b:
## Describe the bug

Attempting to save zstd-encoded content in wayback results in corrupted data being served during replay.

## Steps to reproduce the bug

- Run `wayback --live`.
- Open Firefox.
- Type "http://localhost:8080/" in the address bar.
- Press enter.
- In the page that opens, click on the "/live" link.
- In the URL input, enter "https://goughlui.com/" (I'm not the owner of the blog, it just seems to be one of the few web sites that use zstd).
- Click on the "Search" button.
- Open the first capture that appears in the calendar view.

## Expected behavior

A page identical to the one available through a browser should be served.

## Screenshots

![Corrupted output when replaying zstd-encoded content](https://github.com/user-attachments/assets/ecbf6f46-be9b-4d88-b371-6021a5612baa)

## Environment

- OS: Debian 12
- Browser: Firefox
- Version: 141.0

## Additional context

In order to make sure that the issue was caused by the zstd encoding, I tried removing the zstd entry from Firefox's network.http.accept-encoding.secure setting, and this did indeed make wayback work as expected. One important thing to keep in mind is that the issue can not be reproduced when wayback is not bound to localhost. This is because Firefox will only accept zstd in a "secure" context, and since the HTTP server provided by Werkzeug does not use HTTPS, Firefox will not treat it as a "secure" context unless it is bound to localhost.

PR #948, opened 2025-07-23
Fix issue #947

0x1b5b:
## Description

This pull request fixes issue #947. Since warcio appears to be lacking support for zstd at the moment, I have decided to remove zstd from the Accept-Encoding HTTP header's value. This is similar to the patch that was provided in pull request #444 as a solution to issue #434.

## Motivation and Context

See issue #947.

goughlui.com being served in wayback

Compare this screenshot to the one provided in issue #947.

## Types of changes

## Checklist:

- [ ] I have added or updated tests to cover my changes. (I could not find existing tests for this code path, so I decided not to add any new tests.)
- [ ] All new and existing tests passed. (They do not seem to pass even without my changes.)

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -2,15 +2,12 @@
 
 set -e
 
-# Get UID/GID from volume dir
-VOLUME_UID=$(stat -c '%u' $VOLUME_DIR)
-VOLUME_GID=$(stat -c '%g' $VOLUME_DIR)
-
-MY_UID=$(id -u)
-MY_GID=$(id -g)
-
 # Run as custom user
-if [ "$MY_GID" != "$VOLUME_GID" ] || [ "$MY_UID" != "$VOLUME_UID" ]; then
+if ! [ -w $VOLUME_DIR ]; then
+    # Get UID/GID from volume dir
+    VOLUME_UID=$(stat -c '%u' $VOLUME_DIR)
+    VOLUME_GID=$(stat -c '%g' $VOLUME_DIR)
+
     # create or modify user and group to match expected uid/gid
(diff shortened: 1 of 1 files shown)


pywb/rewrite/rewriteinputreq.py
```diff
@@ -93,11 +93,13 @@ def get_req_headers(self):
                 if self.splits:
                     value = self.splits.scheme
 
-            elif not has_brotli and name == 'HTTP_ACCEPT_ENCODING' and 'br' in value:
+            elif name == 'HTTP_ACCEPT_ENCODING':
                 # if brotli not available, remove 'br' from accept-encoding to avoid
                 # capture brotli encoded content
+                # We have to remove zstd from the list of accepted encodings as warcio does not support it.
+                disallowed_encodings = ('zstd') if has_brotli else ('zstd', 'br')
```
ato, line 100: I think this needs a comma:

```suggestion
                disallowed_encodings = ('zstd',) if has_brotli else ('zstd', 'br')
```

Otherwise if `has_brotli` is true, then `disallowed_encodings` is the string `"zstd"` not a tuple which means the `not in disallowed_encodings` is checking for any substring of 'zstd'.

```
>>> 'z' not in ('zstd')
False
>>> 'z' not in ('zstd',)
True
```

ato (changes requested):

pywb/rewrite/rewriteinputreq.py
wumpus: Good catch

ikreymer (approved):
Makes sense, thanks!

ikreymer:
Confirmed that recording now works for sites that default to zstd, eg. `http://localhost:8080/test/record/https://www.facebook.com/`

PR #949, opened 2025-07-31
Add date range support to access control rules

### LINKED ISSUE #703: Add date ranges to access control lists
## Is your feature request related to a problem? Please describe.

We would like to refine our URL blocks, by specifying a time range for the block, we we can limit access to a subset of the snapshots of URLs, rather than blocking the whole URL for all time.

## Describe the solution you'd like

Support for a syntax like the `embargo` syntax, e.g. these two statements (which would have similar effects!):

```
org,httpbin)/ - {"access": "block", "url": "httpbin.org/", "before": "20201226"}
org,httpbin)/anything/something - {"access": "allow", "url": "http://httpbin.org/anything/something", "after":"202001226}

```

## Describe alternatives you've considered

We could delete CDX records but we don't want to block the same URLs across all access contexts.

tw4l:
## Description

Fixes #703 

This PR adds date range support to ACL rules that uses the existing embargo syntax.

Access can be controlled by specifying `before` or `after` a particular timestamp, e.g.

```
org,iana)/ - {"access": "allow", "url": "http://www.iana.org/", "before": "20140126"}
```

Time intervals can also be used via `newer` and `older`, e.g.:

```
org,iana)/ - {"access": "allow", "url": "http://www.iana.org/", "newer": {"years": 1, "months": 6}}
```

Documentation has also been added.

## Motivation and Context

This will allow users to block/allow a subset of captures from a URL based on date, e.g. in the case where ownership of a URL changed at some point.

## Types of changes

## Checklist:

- [X] I have added or updated tests to cover my changes.

### DIFF
--- a/tests/test_acl.py
+++ b/tests/test_acl.py
@@ -109,10 +109,10 @@ class TestACLApp(BaseConfigTest):
 
         resp = self.testapp.get('/pywb-acl-before/20140126200624mp_/http://www.iana.org/', status=200)
 
-        resp = self.testapp.get('/pywb-acl-before/20140126200825/http://www.iana.org/domains', status=451)
+        resp = self.testapp.get('/pywb-acl-before/20140126200825mp_/http://www.iana.org/domains', status=451)
         assert 'Access Blocked' in resp.text
 
-        resp = self.testapp.get('/pywb-acl-before/20140126201248/http://www.iana.org/domains/arpa', status=200)
+        resp = self.testapp.get('/pywb-acl-before/20140126201248mp_/http://www.iana.org/domains/arpa', status=200)
 
     def test_acl_after(self):
         resp = self.testapp.get('/pywb-acl-after/20140126200624mp_/http://www.iana.org/', status=451)


docs/manual/access-control.rst
```diff
@@ -166,6 +166,41 @@ Further examples of how to set this header will be provided in the deployments s
...
+Access can also be controlled by specifying a relative time interval, similar to embargos.
+
+For example, the following access control settings restrict access to ``https://example.com/restricted/`` by default, but allow access to all captures newer than 1 year::
+
+  com,example)/restricted - {"access": "allow", "older": {"years": 1}}
+  com,example)/restricted - {"access": "block"}
+
+The following access control settings restrict access to ``https://example.com/restricted/`` by default, but allow access to all captures older than 1 year, 2 months, 3 weeks, and 4 days::
+
+  com,example)/restricted - {"access": "allow", "older": {"years": 1}, "months": 2, "weeks": 3, "days": 4}
```
tw4l, line 198:
```suggestion
  com,example)/restricted - {"access": "allow", "older": {"years": 1, "months": 2, "weeks": 3, "days": 4}}
```

docs/manual/access-control.rst
```diff
@@ -166,6 +166,41 @@ Further examples of how to set this header will be provided in the deployments s
...
+
+
+Date-Based Access Controls: Time Interval
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Access can also be controlled by specifying a relative time interval, similar to embargos.
+
+For example, the following access control settings restrict access to ``https://example.com/restricted/`` by default, but allow access to all captures newer than 1 year::
+
+  com,example)/restricted - {"access": "allow", "older": {"years": 1}}
```
tw4l, line 193:
```suggestion
  com,example)/restricted - {"access": "allow", "newer": {"years": 1}}
```

ikreymer:
Looks great! Would be good to have a slightly more complex test case, with multiple entries or multiple paths to ensure that path matching is working as expected

Issue #954, opened 2025-09-23
Installation results in Greenlet build error (Linux Mint 22.1 / Python 3.12.3)


bitsgalore:
## Describe the bug

Installation of latest Pywb fails with build error for greenlet, installable package.

## Steps to reproduce the bug

I tried to install Pywb with the following command:

```
pipx install pywb
```

Result:

```
Fatal error from pip prevented installation. Full pip output in file:
    /home/johan/.local/state/pipx/log/cmd_2025-09-23_13.39.59_pip_errors.log

pip seemed to fail to build packages:
    greenlet
    installable

Some possibly relevant errors from pip install:
    error: subprocess-exited-with-error
    src/greenlet/greenlet.h:8:10: fatal error: Python.h: No such file or directory
    error: command '/usr/bin/x86_64-linux-gnu-gcc' failed with exit code 1
    error: failed-wheel-build-for-install

Error installing pywb.
```

See also attached [cmd_2025-09-23_13.39.59_pip_errors.log](https://github.com/user-attachments/files/22493033/cmd_2025-09-23_13.39.59_pip_errors.log) for details.

## Environment

- OS: Linux Mint 22.1 Xia
- Python version:  3.12.3

PR #956, opened 2025-09-24
Python 3.12 support

bitsgalore:
## Description

This PR adds support for Python 3.12. 

## Motivation and Context

Pywb currently fails to install under Python 3.12, see [this issue](https://github.com/webrecorder/pywb/issues/954). I fixed this with the following changes:

- [raised python_requires upper limit to <3.13](https://github.com/bitsgalore/pywb/commit/7450453ce1630a1d00f215a7b299f19d029d36af)
- [updated greenlet dependency to =>3.1.1](https://github.com/bitsgalore/pywb/commit/ba65285492c9d2e4ab0799e21e2385e253c075ba)
- [updated gevent dependency to >=25.0.0](https://github.com/bitsgalore/pywb/commit/452f5c52b9ae2788aede0e59811a32e5e05da23c)

With these changes I was able to successfully install Pywb under Python 3.12 (I'm running Linux Mint 22.1 Xia).

A few things to watch out for:

- I'm not entirely sure that these changes won't break things for older Python versions.
-  Perhaps python_requires could be relaxed even further to include 3.13 (`<3.14`), but I have no idea really!
- I tested this by installation Pywb with my changes locally with pipx, which installs everything (including the dependencies) in a virtual environment. However, to run the the unit tests these dependencies must be available system wide (at least by default). For various reasons I don't want to this global install right now, and I don't know how to use the virtualenv for this. This didn't used to be a problem in older Python versions, but it seems handling of packages changed quite a bit around Python 3.11/3.12, and I'm not entirely sure how this currently works.  

Anyway, feel free to ignore if you don't want to merge these changes because of all this. But hopefully this is of some use.

## Types of changes

- [ x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,7 +8,7 @@ brotlipy
 pyyaml
 werkzeug==2.2.3
 webencodings
-gevent==22.10.2
+gevent>=25.0.0
 greenlet>=3.1.1
 webassets==2.0
 portalocker
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,7 +9,7 @@ pyyaml
 werkzeug==2.2.3
 webencodings
 gevent==22.10.2
-greenlet>=2.0.2,<3.0
+greenlet>=3.1.1
 webassets==2.0
 portalocker
 wsgiprox>=1.5.1
--- a/setup.py
+++ b/setup.py
@@ -126,7 +126,7 @@ setup(
             "translate_toolkit"
         ],
     },
-    python_requires='>=3.7,<3.12',
+    python_requires='>=3.7,<3.13',
     tests_require=load_requirements("test_requirements.txt"),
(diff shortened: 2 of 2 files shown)


tw4l:
Thanks @bitsgalore ! Really appreciate this PR. The first blocker I see to merging this (the main thing that had kept us from upgrading previously) is the httpbin dependencies of `greenlet<3.0; python_version < "3.12"`. As many of the tests rely on httpbin, we will likely need to either fork that project from https://github.com/psf/httpbin or find a suitable replacement/alternative in the tests so that we don't end up losing a fair amount of test coverage.

bitsgalore:
@tw4l Thanks for the heads up, great to see this fixed!

PR #959, opened 2025-10-03
Better fix for #865 / differentiate between remote cdx + local WARCs and remote cdx + liveweb loader

### LINKED ISSUE #865: Pywb failing to handle self-redirects from OutbackCDX

ikreymer:
The fix is to be more concretely differentiate between these two configurations:

```
# remote cdx + remote proxy loading via LiveWebLoader
remote: cdx+https://remote-archive.example.com/web/cdx

# remote cdx + but *local* WARC loading:
local_coll:
    index_paths: cdx+http://outbackcdx:8080/coll
    archive_paths: /path/to/mywarcs/
```

The former is for proxying via remote CDX while the latter is the [recommend config for OutbackCDX](https://pywb.readthedocs.io/en/latest/manual/outbackcdx.html#configure-pywb-with-outbackcdx) 

These supersedes the fix proposed in #917 by adding a conditional to check which configuration is used
(remote cdx + local loading or remote cdx + remote liveweb loading) by checking if `archive_paths` was provided,
as is needed for usage with OutbackCDX

The skipping of the livewebloader
is needed to be able to retry the lookup path again for revisit records, while of course remote proxying
will not need to do that.

Fixes #865 

## Types of changes

## Checklist:

### DIFF
--- a/pywb/warcserver/handlers.py
+++ b/pywb/warcserver/handlers.py
@@ -175,9 +175,10 @@ class ResourceHandler(IndexHandler):
 #=============================================================================
 class DefaultResourceHandler(ResourceHandler):
     def __init__(self, index_source, warc_paths='', forward_proxy_prefix='',
+                 use_local_file_load=False,
                  **kwargs):
         loaders = [WARCPathLoader(warc_paths, index_source),
-                   LiveWebLoader(forward_proxy_prefix),
+                   LiveWebLoader(forward_proxy_prefix, use_local_file_load=use_local_file_load),
                    VideoLoader()
                   ]
         super(DefaultResourceHandler, self).__init__(index_source, loaders, **kwargs)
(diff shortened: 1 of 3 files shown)


tw4l (approved):
Makes sense and testing well!

PR #967, opened 2025-11-03
Add inject_scripts option

ato:
## Description

This adds an option to inject custom scripts from static/ into replayed pages in both the client-side and server-side replay modes.

```yaml
inject_scripts:
  - ruffle/ruffle.js
  - tweaks.js
```

## Motivation and Context

This is useful for emulating removed browser features, and applying compatibility or behavior tweaks. For example:

- injecting [Ruffle](https://ruffle.rs/) to emulate Flash Player
- setting `document.layers = true` on a specific site to bypass "You must use Netscape 4.x"
- disabling (or warning on) form submission to prevent user confusion

We've been injecting scripts by overriding `head_insert.html` (for server-side replay mode) and `loadWabac.js` (for client-side replay mode) but it would be cleaner to have a single option that works for both modes. It would also be nice to not have to repatch the templates each time we update pywb.

## Types of changes

## Checklist:

- [ ] All new and existing tests passed. (*There's a CI test failure but as far as I can tell it's unrelated to this change*)

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -17,6 +17,7 @@ python =
 setenv = PYWB_NO_VERIFY_SSL = 1
 passenv = *
 deps =
+    setuptools
     -rtest_requirements.txt
     -rrequirements.txt
     -rextra_requirements.txt


tests/config_test.yaml
```diff
@@ -41,3 +41,6 @@ enable_memento: true
 
 # enable new transclusion system
 transclusions_version: 2
+
+inject_scripts:
+  - inject.js
```
tw4l:
```suggestion
  - inject.js

```

Add newline

pywb/static/loadWabac.js
```diff
@@ -54,6 +55,7 @@ class WabacReplay
         archiveMod: "ir_",
         adblockUrl: this.adblockUrl,
         noPostToGet: true,
+        injectScripts: this.injectScripts.map(src => "../" + src),
```
tw4l: pywb supports static files from the root static directory or per-collection static directories: https://pywb.readthedocs.io/en/latest/manual/ui-guide.html#static-files. It might be worth applying that same logic to the injected scripts, in case users want e.g. Ruffle enabled but only on one collection?

tw4l (commented):
Thanks for the submission! I suggested a few in-line changes, and have one question around whether we should also support injecting scripts from the per-collection static directories.

pywb/static/loadWabac.js
ato: Oh and the `injectScripts.map(src => "../" + src)` on this line is because wabac.js actually adds the prefix `/static/proxy/`. It looked to me like maybe the original idea was for it to load absolute URLs from the live web, but the serviceworker just returned 404 when I tried that, but `../` worked.

ato:
Oh, my bad, thanks for catching that Tessa. And thanks for for the fix Lauren.

> In the case of `live` or a `remote` one, perhaps the inject_scripts on a per collection basis doesn't make sense?

Yeah, I guess you can at least use the top-level setting for them and in a pinch you could work around it by having the injected JavaScript file itself check the collection name.

To allow for per-collection options though it would probably make sense if '$live' and '$all' were just shorthand for something like:

```yaml
collections:
  all:
     type: all
  live:
     type: live
``` 

but that's probably better tackled separately to this.

> I can't recall how things work if you have collections defined in config.yaml and one by the same name in the `collections` directory.

It seems ok? At least I don't notice anything obviously wrong when I setup the directories under `collections` and only define `inject_scripts`:

```yaml
collections:
  test1:
    inject_scripts:
      - test1.js
  test2:
    inject_scripts:
      - test2.js
```

ikreymer (approved):
Nice work! Seems like this has been reviewed and tested. Great that inject scripts can be both global and per collection, matching existing convention. Thanks for updating the docs as well!

PR #968, opened 2025-11-05
Unregister the service worker when client_side_replay is off

ato:
## Description

This change makes pywb unregister the wabac.js serviceworker when `client_side_replay: false` is set.

## Motivation and Context

If you have client_side_replay on and later turn it off, prior visitors will still have it registered in their browser unless they manually remove it through devtools or by clearing site data. While we don't expect to be regularly flipping it on and off in production, it's important to be able to roll back to the previous configuration if an issue is discovered.

## Types of changes

## Checklist:

### DIFF
--- a/pywb/templates/frame_insert.html
+++ b/pywb/templates/frame_insert.html
@@ -53,12 +53,14 @@ html, body
 <script>
   window.cframe = null;
 
-{% if client_side_replay %}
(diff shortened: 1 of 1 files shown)


ikreymer (approved):
Thanks for adding this - Might require another reload, but not sure there's another way.

PR #974, opened 2025-11-27
deps: bump wabac.js to 2.25.0 to fix #973

### LINKED ISSUE #973: JS static field initialiser called 'location' incorrectly rewritten
## Describe the bug
When the recorded website contains a JavaScript class with a static field initialiser called `location = X`, it is mapped to `location = ((self.__WB_check_loc && self.__WB_check_loc(location, arguments)) || {}).href = X`. Since `arguments` is invalid in this context, a `SyntaxError` is immediately thrown.

This is particularly concerning because as of Angular 19 (which came with the introduction of their new build tool `@angular-devkit/build-angular:application`), this code is injected. That means that any page running on Angular 19 (or up) can no longer be recorded.

## Steps to reproduce the bug

Input:
```
class X {
    location = "X";
};
```

Output:
```
class X {
    location = ((self.__WB_check_loc && self.__WB_check_loc(location, arguments)) || {}).href = "X";
};
```

Javascript:
firefox: `Uncaught SyntaxError: arguments is not valid in fields`
chrome: `Uncaught SyntaxError: 'arguments' is not allowed in class field initializer or static initialization block`

## Expected behavior
Since this is a static field that happens to be called `location`, which has nothing to do with the "actual" `location` or `window.location`, I expect this code to be untouched. In case it needs to be rewritten, it needs to not use `arguments`.

## Screenshots
-

## Environment
Browser: chrome/firefox
Angular version: 19
pywb version: 2.9.1

ikreymer:
version: bump to 2.10.0b1

fixes #973 via wabac.js update

### DIFF
--- a/pywb/version.py
+++ b/pywb/version.py
@@ -1,4 +1,4 @@
-__version__ = '2.10.0b0'
+__version__ = '2.10.0b1'
 
 if __name__ == '__main__':
     print(__version__)
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ from pywb import __version__
 root_dir = pathlib.Path(__file__).parent
 
 
-WABAC_SW_URL = "https://cdn.jsdelivr.net/npm/@webrecorder/wabac@2.24.1/dist/sw.js"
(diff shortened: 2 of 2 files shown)


ikreymer:
Just tagging for testing the upcoming beta so can do a 2.10.0 release!

PR #991, opened 2026-04-07
Python 3.13 and 3.14 support

### LINKED ISSUE #979: Can't install package in python 3.13+ enviornments
## Describe the bug

I get this error whenever I try to install in a python environment 3.13+ or higher:

```
failed to build `gevent==23.9.0.post1` ├─▶ the build backend returned an error ╰─▶ call to `setuptools.build_meta.build_wheel` failed (exit status: 1)
```
## Steps to reproduce the bug

 These commands errored out:

```
uv add pywb -p .venv/bin/python3.13
uv add pywb -p .venv_3_14/bin/python3.14
```

Same for python 3.15; confirmed dev tools are installed on my machine.

## Expected behavior

Package should install without error or document 3.13+ not supported

## Environment

On mac

ato:
## Description

- update werkzeug, gevent, greenlet to latest version for 3.14 support but conditionally pin when necessary to not break older Python
- Flask 2.2.0 replaced JSONIFY_PRETTYPRINT_REGULAR with app.json.compact so support both
- setuptools is no longer included by default since Python 3.12 so add it
- cgi was removed in Python 3.13 so replace it with legacy-cgi
- add python 3.13 and 3.14 to test matrix in both tox and the Github action

## Motivation and Context

To allow pywb to run on Python 3.13 and 3.14.
Fixes #979

## Types of changes

## Checklist:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,12 +7,10 @@ jinja2>=3.1.2
 surt>=0.3.1
 brotlipy
 pyyaml
-werkzeug==2.2.3; python_version<"3.9"
-werkzeug==3.1.7; python_version>="3.9"
+werkzeug==3.1.7
 webencodings
 legacy-cgi; python_version>="3.13"
-gevent==22.10.2; python_version<"3.8"
-gevent==23.9.0.post1; python_version>="3.8" and python_version<"3.12"
+gevent==23.9.0.post1; python_version<"3.12"
 gevent==25.9.1; python_version>="3.12"
 greenlet>=2.0.2,<3.0; python_version<"3.12"
 greenlet==3.3.2; python_version>="3.12"
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,14 +7,15 @@ jinja2>=3.1.2
 surt>=0.3.1
 brotlipy
 pyyaml
-werkzeug==2.2.3
+werkzeug==2.2.3; python_version<"3.9"
+werkzeug==3.1.7; python_version>="3.9"
 webencodings
 legacy-cgi; python_version>="3.13"
 gevent==22.10.2; python_version<"3.8"
-gevent==23.9.0.post1; python_version>="3.8" and python_version<"3.13"
-gevent==24.10.1; python_version>="3.13"
+gevent==23.9.0.post1; python_version>="3.8" and python_version<"3.12"
+gevent==25.9.1; python_version>="3.12"
 greenlet>=2.0.2,<3.0; python_version<"3.12"
-greenlet==3.2.4; python_version>="3.12.0rc0"
+greenlet==3.3.2; python_version>="3.12"
 webassets==2.0
 portalocker
 wsgiprox>=1.5.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
 six
+setuptools; python_version>="3.12"
 warcio>=1.7.5
 requests
 redis==2.10.6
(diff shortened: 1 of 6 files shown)


setup.py
```diff
@@ -127,7 +127,7 @@ def get_package_data():
             "babel-vue-extractor"
         ],
     },
-    python_requires='>=3.7,<3.13',
+    python_requires='>=3.7,<3.15',
```
ldko: Trying to run pywb in 3.8 results in `ImportError: cannot import name 'files' from 'importlib.resources'`, presumably introduced via the [replacement of pkg_resources](https://github.com/webrecorder/pywb/commit/00df79d196bcabdd77993dd7e8d7753fc7b17bde), since [`files` was not available until 3.9](https://docs.python.org/3/library/importlib.resources.html#importlib.resources.files). Since Python 3.8 has been end-of-life since 2024-10-07, it would make sense to drop pretense of supporting it and earlier.
```suggestion
    python_requires='>=3.9,<3.15',
```

tox.ini
```diff
@@ -4,14 +4,16 @@ testpaths =
     tests
 
 [tox]
-envlist = py37, py38, py39, py310, py311, py312
+envlist = py37, py38, py39, py310, py311, py312, py313, py314
```
ldko:
```suggestion
envlist = py39, py310, py311, py312, py313, py314
```

setup.py
ldko: We should take out these lines mentioning 3.7 and 3.8 unless there is a consensus to further fix things for 3.7 and 3.8 compatibility.

requirements.txt
```diff
@@ -1,17 +1,21 @@
...
 warcio>=1.7.5
 requests
 redis==2.10.6
 jinja2>=3.1.2
 surt>=0.3.1
 brotlipy
 pyyaml
-werkzeug==2.2.3
+werkzeug==2.2.3; python_version<"3.9"
+werkzeug==3.1.7; python_version>="3.9"
```
ldko: It might make sense to simplify this if pywb is not compatible with less than 3.9.
```suggestion
werkzeug==3.1.7
```

requirements.txt
```diff
@@ -1,17 +1,21 @@
...
 webencodings
+legacy-cgi; python_version>="3.13"
 gevent==22.10.2; python_version<"3.8"
-gevent==23.9.0.post1; python_version>="3.8"
+gevent==23.9.0.post1; python_version>="3.8" and python_version<"3.12"
```
ldko:
```suggestion
gevent==23.9.0.post1; python_version>="3.9" and python_version<"3.12"
```

requirements.txt
```diff
@@ -1,17 +1,21 @@
...
 jinja2>=3.1.2
 surt>=0.3.1
 brotlipy
 pyyaml
-werkzeug==2.2.3
+werkzeug==2.2.3; python_version<"3.9"
+werkzeug==3.1.7; python_version>="3.9"
 webencodings
+legacy-cgi; python_version>="3.13"
 gevent==22.10.2; python_version<"3.8"
```
ldko: This line is no longer relevant.

ldko (commented):
The changes, including suggestions related to dropping 3.7 and 3.8 mentions, worked for me running pywb in 3.9-3.14.

ato:
Thanks for testing on 3.8. I agree with you on dropping 3.7 and 3.8 rather than fixing. They don't even seem very relevant for long support Linux:

Distribution | Default Python | End of Support
-- | -- | --
Debian 10 | **Python 3.7** | 2024
Debian 11 | Python 3.9 | 2026
RHEL 8 | Python 3.6 | 2029
RHEL 9 | Python 3.9 | 2032
SLES 15 | Python 3.6 | 2031
SLES 16 | Python 3.13 | 2035
Ubuntu 20.04 | **Python 3.8** | 2025; 2031 for Ubuntu Pro
Ubuntu 22.04 | Python 3.10 | 2027

wumpus:
We should test all of the python versions we claim to support. I think they're aligned now. `warcio` and `cdx-toolkit` are also currently aligned.

ldko:
Not sure about the Cancelled test on python 3.9 that has led to a failing check. Tests passed locally for me though. Thanks for adding 3.13 and 3.14 support @ato

ldko (approved):

ato:
> Not sure about the Cancelled test on python 3.9 that has led to a failing check.

The last line logged before hitting the 15m timeout was `tests/test_acl.py::TestACLApp::test_acl_after PASSED` which seems an unlikely place to get stuck. Perhaps an issue with the test host. It passed when I clicked retry.

PR #1002, opened 2026-08-26
Fix empty replayPrefix in client-side replay service worker registration

klaegera:
## Description

Fix `loadWabac.js` to pass the actual replay prefix when registering the service worker for client-side replay.

Based on the surrounding code in `loadWabac.js`, it seems likely that this was the original intention.

## Motivation and Context

We experienced various strange issues involving pages not loading properly in client replay mode when serving via one or more reverse proxies (traefik + oauth2-proxy in our case). The issues do not appear when serving pywb directly from localhost.

While debugging, we found that WabacReplay was registering the service worker with an empty `replayPrefix` query parameter:
```
/static/sw.js?replayPrefix=
```

Providing the correct replay prefix to the service worker in `loadWabac.js` resolved the issue for us.

## Types of changes

## Checklist:

### DIFF
--- a/pywb/static/loadWabac.js
+++ b/pywb/static/loadWabac.js
@@ -11,7 +11,7 @@ class WabacReplay
     this.injectScripts = injectScripts;
     this.adblockUrl = undefined;
 
-    this.queryParams = {"replayPrefix": ""};
+    this.queryParams = {"replayPrefix": this.prefix};
     if (this.isRoot) {
       this.queryParams["root"] = "$root";
     }


ato:
Within wabac.js, `replayPrefix` is prepended with `prefix` here:

https://github.com/webrecorder/wabac.js/blob/61ddc35238ba87bdf562e20619f678a10e457f92/src/swmain.ts#L239-L241

Unless I'm misunderstanding something, this change would therefore result in it being doubled up, something like `https://host/coll/https://host/coll//`. I believe that setting it to the empty string is intentional (the default value is "w") so that URLs handled by the serviceworker are the same as normal server-side replay mode URLs.

klaegera:
Thank you for you feedback. I think you're right.

After more experimentation, my current theory is that our proxies 'sanitizing' the double slashes in pywb's URLs is causing the problem and setting the `replayPrefix` acted as a workaround in our case for other reasons.

PR #1005, opened 2026-09-17
Relax dependencies about gevent and greenlet

amotl:
Hi there. Thanks a stack for conceiving and maintaining pywb. We observed problems around gevent when trying to install the package on macOS/Python 3.14. Relaxing corresponding dependency constraints solved it for us.


ato:
Thanks! That's my bad, I'm too used to pinning specific versions from Maven.

I did some testing and fully unconstrained gevent with greenlet pulled in transitively seems to work with all the Python versions we support, so I ended up going with just bare `gevent` in 872e21d55a37519c85d08fa99c0485d58b08963d.

wumpus:
I am pleasantly surprised that this works -- last time @tw4l wrestled with gevent / greenlet, it was complicated. But the thing @ato pushed passed Ci, so...

ato:
I was surprised too! I think dropping Python 3.7 and 3.8 support simplified things.

